PR #126, opened 2014-11-27
dump errors & debugging output to stderr

bjaglin:
I tried to go through the places where I thought it made sense to dump output to stderr. This is just an initial take at it - more of a proposal.


bjaglin:
any feedback on this?

michaelsauter:
Sorry, forgot about this one. Could you explain why you want to change this? Looks a bit odd to me to move e.g. the version output to STDERR.

bjaglin:
Maybe the change related to the version is the one that makes less sense, I'll remove it. But for the others, I believe it's common to separate the canonical output from errors and debug messages, for example when you pipe the output to a file but want to keep logs in your shell.

michaelsauter:
Ah okay. Makes sense to me. I look over it and include it in the next minor release.

Issue #120, opened 2014-11-04
Add support for automatically creating a 'latest' tag for `crane push` and `crane provision`


dancrumb:
It would be nice to have the ability to provision and push images with the `latest` tag via a CLI parameter.

I already version my images numerically, but it's convenient to have a 'moving' tag that points to the latest version, but doing that with crane is not a trivial task.

I'd like something like

`crane provision --latest && crane push --latest`

michaelsauter:
Sorry for the long silence on this one. I think overall, providing `--tag` is useful in a few cases and doesn't hurt people who don't need it. Somehow the whole thing doesn't feel a 100% right, but I'm unsure what would improve the situation under the current approach (which won't change since we're post 1.0), so I think this is the best way to go.

michaelsauter:
Closing since there won't be any more build-related features in Crane, see #195.

PR #131, opened 2014-12-17
Issue #120 Add support for --tag to lift, provision and push

dancrumb:
Sorry about all the back and forth this evening. Git and I had a falling out

This Pull Request should be everything we need.


.gitignore
```diff
@@ -1,2 +1,10 @@
 crane_darwin_amd64
 crane_linux_amd64
+
+# IntelliJ IDE files
+.idea/
+atlassian-ide-plugin.xml
+*.iml
+
+# Cloud9 IDE files
+.c9/
```
michaelsauter: I'm not a huge fan of putting all the different project files into the `.gitignore` file. My suggestion would be to remove them here and put them in your global `.gitignore`.

README.md
```diff
@@ -43,9 +43,10 @@ Maps to `docker unpause`.
 
 ### `provision`
 Either calls Docker's `build` or `pull`, depending on whether a Dockerfile is specified. The Docker cache can be disabled by passing `--no-cache`.
+A specific tag can be build using the `--tag` argument;
 
 ### `push`
-Maps to `docker push`.
+Maps to `docker push`. By using the `--tag` parameter, you can push your image with a specific tag, instead of the version that is in your Crane config file.
```
michaelsauter: Maybe more precise would be `instead of the version that may be specified in your config file`.

crane/cmd.go
```diff
@@ -227,11 +228,15 @@ See the corresponding docker commands for more information.`,
...
 	cmdLift.Flags().BoolVarP(&options.recreate, "recreate", "r", false, "Recreate containers (force-remove containers if they exist, force-provision images, run containers)")
 	cmdLift.Flags().BoolVarP(&options.nocache, "no-cache", "n", false, "Build the image without any cache")
+	cmdLift.Flags().StringVarP(&options.tag, "tag", "t", "", "Use a specific tag when building images (overriding any set in the config file)")
 
 	cmdProvision.Flags().BoolVarP(&options.nocache, "no-cache", "n", false, "Build the image without any cache")
+	cmdProvision.Flags().StringVarP(&options.tag, "tag", "t", "", "Use a specific tag when building images (overriding any set in the config file)")
 
 	cmdCreate.Flags().BoolVarP(&options.recreate, "recreate", "r", false, "Recreate containers (force-remove containers first)")
 
+	cmdPush.Flags().StringVarP(&options.tag, "tag", "t", "", "Push to a specific tag (overriding any set in the config file)")
```
michaelsauter: Should this be just `Push a specific tag ...`?

crane/container.go
```diff
@@ -639,16 +639,29 @@ func (c *container) Logs(follow bool, tail string) (stdout, stderr io.Reader) {
...
 		print.Noticef("Skipping %s as it does not have an image name.\n", c.Name())
 	}
 }
 
+func (c *container) getTaggedImageName(tag string) {
+	imageName := c.Image()
+	if len(imageName) > 0 {
+		if len(tag) > 0 {
+			imageParts := strings.Split(imageName, ":")
+			imageParts[1] = tag
```
michaelsauter: Agreed, I don't see a problem with that.

michaelsauter:
Looks good to me so far, haven't had the chance to give it a spin.

I myself use SublimeText, which works well enough for me. Not sure how exactly I started out originally, but I do think the official docs are actually pretty good, though maybe somewhat different to other tutorials/docs. [How to Write Go Code](https://golang.org/doc/code.html) followed by [Effective Go](https://golang.org/doc/effective_go.html) should get you started.

dancrumb:
Thanks; I'll check those out. I agree with your recommendations and have made the relevant changes.

crane/container.go
bjaglin: Maybe I am missing something, but I just wanted to point out that this look like it will panic if the result of the split is a single-item array.

dancrumb: You're right; I spend most of my line with languages that support array insertion via this method. I'll fix this today.

.gitignore
```diff
@@ -1,2 +1,2 @@
 crane_darwin_amd64
-crane_linux_amd64
+crane_linux_amd64
```
michaelsauter: This file should not be changed, and contain a newline at the end.

crane/container.go
```diff
@@ -639,16 +639,36 @@ func (c *container) Logs(follow bool, tail string) (stdout, stderr io.Reader) {
...
+		args := []string{"push", imageName}
 		executeCommand("docker", args)
 	} else {
 		print.Noticef("Skipping %s as it does not have an image name.\n", c.Name())
 	}
 }
 
+func (c *container) nameWithTagOverride(tag string) string {
+	imageName := c.Image()
+	fmt.Println(imageName)
```
michaelsauter: I don't think this should be printed here. Also below. I don't think anything should be printed in this method.

michaelsauter:
Closing this due to inactivity. Still open for the feature request though if someone wants to take over or start again :)

PR #133, opened 2015-01-07
Add forwarded port to the output

partkyle:
This is just a quick hack to add the forwarded port to the crane status.

This makes the output look like the following:

```
NAME      IMAGE       ID            UP TO DATE  IP            PORTS                     RUNNING
memcached memcached   2a135b55e234  false       172.17.0.65   11211/tcp,                true
mysql     mysql:5.5   64d4f8652df2  true        172.17.0.63   3306/tcp,                 true
apid      web         b7f0dec86978  true        172.17.0.66   0.0.0.0:18082->8082/tcp,  true
```

### DIFF
--- a/crane/container.go
+++ b/crane/container.go
@@ -8,7 +8,6 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/flynn/go-shlex"
 	"github.com/michaelsauter/crane/print"
 )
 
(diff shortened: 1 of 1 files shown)


crane/container.go
```diff
@@ -2,13 +2,14 @@ package crane
...
-	"github.com/flynn/go-shlex"
-	"github.com/michaelsauter/crane/print"
 	"io"
 	"os"
 	"path"
 	"strconv"
 	"strings"
+
+	"github.com/flynn/go-shlex"
+	"github.com/michaelsauter/crane/print"
```
michaelsauter: Did you change the order? The order of the imported packages is determined by `gofmt` and should not be changed. Or has `gofmt` output changed and I'm unaware of that?

crane/container.go
```diff
@@ -345,7 +346,7 @@ func (c *container) ImageExists() bool {
 
 func (c *container) Status() []string {
 	fields := []string{c.Name(), c.Image(), "-", "-", "-", "-", "-"}
-	output := inspectString(c.Id(), "{{.Id}}\t{{.Image}}\t{{if .NetworkSettings.IPAddress}}{{.NetworkSettings.IPAddress}}{{else}}-{{end}}\t{{range $k,$v := $.NetworkSettings.Ports}}{{$k}},{{else}}-{{end}}\t{{.State.Running}}")
+	output := inspectString(c.Id(), "{{.Id}}\t{{.Image}}\t{{if .NetworkSettings.IPAddress}}{{.NetworkSettings.IPAddress}}{{else}}-{{end}}\t{{range $networkport,$hostforwardmap := $.NetworkSettings.Ports}}{{range $_, $hostinfo := $hostforwardmap}}{{$hostinfo.HostIp}}:{{$hostinfo.HostPort}}->{{$networkport}},{{else}}{{$networkport}},{{end}}{{else}}-{{end}}\t{{.State.Running}}")
```
michaelsauter: Please use camelCase consistently, e.g. `$networkPort` and `$hostForwardMap` etc.

crane/container.go
partkyle: I'm using goimports, so maybe that's why. It's always separated the remote packages. I can change this if you prefer it.

crane/container.go
partkyle: fixed.

crane/container.go
michaelsauter: Interesting. Anyway, the pull request now removes `go-shlex` which is definitely needed, so this cannot compile. Please fix (and ideally I'd like the import statements not to change at all, since they're not related to this PR).

michaelsauter: @partkyle ping

PR #136, opened 2015-01-24
Better validation/error handling for targets

bjaglin:
We had a typo in one of the containers declared in a group, and it took us hours to troubleshoot the issue... This ensures that we get a verbose failure when such a group is targeted.

I though about implementing systematic validation for all groups, but I think it's a bad idea because some groups might have valid references only when some environment variables are set, and these variables or there values could differ from one group to another.


michaelsauter:
I like this a lot! In general, one area where Crane can definitely be improved is more error reporting etc. Unfortunately, I don't have any time to work on it right now, but hopefully that will change :)

Thanks again for this!

PR #139, opened 2015-01-27
Add mac-address support (Docker 1.4)

lefeverd:
Hello,
since Docker 1.4 it is possible to add a --mac-address parameter to the run command.

I don't really know about what version of Docker you want to support but as I needed this parameter, I suggest this pull request.

Regards,
David

### DIFF
--- a/crane/config.go
+++ b/crane/config.go
@@ -228,49 +228,67 @@ func (c *config) determineTarget(target []string, cascadeDependencies string, ca
 // explicitlyTargeted receives a target and determines which
 // containers of the map are targeted.
 func (c *config) explicitlyTargeted(target []string) (result []string) {
-	result = []string{}
-	// target not given
 	if len(target) == 0 {
-		// If default group exists, return its containers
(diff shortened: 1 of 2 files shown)


crane/container.go
```diff
@@ -78,6 +79,7 @@ type RunParameters struct {
 	RawVolumesFrom []string    `json:"volumes-from" yaml:"volumes-from"`
 	RawWorkdir     string      `json:"workdir" yaml:"workdir"`
 	RawCmd         interface{} `json:"cmd" yaml:"cmd"`
+
 }
```
michaelsauter: Please remove the blank line here.

michaelsauter:
Thanks for this! Definitely want to merge it. Crane currently depends on Docker >= 1.3, but I think it's fine to add more params and just add a note to the readme as you did. Could I ask you to squash into one commit? BTW, you're second commit looks like it's from a different author, just make sure you've got the author set correctly so that you're listed in contributors file :)

PR #144, opened 2015-02-01
Docker 1.5 flags

bjaglin:
Tested against Docker 1.5.0-rc2

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ The map of containers consists of the name of the container mapped to the contai
 	* `lxc-conf` (array)
 	* `mac-address` (string) Need Docker >= 1.4
 	* `memory` (string)
+	* `memory-swap` (string) Need Docker >= 1.5
 	* `net` (string) The `container:id` syntax is not supported, use `container:name` if you want to reuse another container network stack.
 	* `privileged` (boolean)
 	* `publish` (array) Map network ports to the container.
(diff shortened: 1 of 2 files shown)


dreamcat4:
+1. I need this too.

dreamcat4:
@bjaglin - Did I do it right? I've tried to put `pid: host` in the run options of crane.yml, but it hasn't seemed to work. My config file looks like this:

```
  pipework:
    image: dreamcat4/pipework
    dockerfile: pipework
    run:
      volume:
        - /var/run/docker.sock:/docker.sock
      privileged: true
      pid: host
      link:
        - tvheadend:tvheadend
```

I've already installed docker-1.5.0 from the ubuntu testing repository. Confirmed that same image works otherwise on cmd line with `docker run --pid=host …`. Where it sees the host processes.

Did you test new `--pid` option specifically, or just the other 2 ones?

dreamcat4:
@bjaglin ah, never mind. Checked your code which looked OK. It seems I didn't compile locally properly. It pulled in the official version and dumped it in src/ subdirectory…

If you could please give me a better build instructions. What I did previously (which has gotta be wrong somewhere) was this:

	sudo apt-get install gccgo-go
	git clone https://github.com/bjaglin/crane.git crane-docker15flags
	cd crane-docker15flags
	git checkout -b docker15flags origin/docker15flags
	GOPATH=$PWD go get # ignore the error message
	GOPATH=$PWD go build
	./crane-docker15flags
	ln -s $PWD/crane-docker15flags /usr/local/sbin/crane-docker15flags

michaelsauter:
@bjaglin Thanks again!

@dreamcat4 Merged into master, so this should be a lot easier to build :)

PR #145, opened 2015-02-03
add mapping for missing run flags

bjaglin:


crane/container.go
```diff
@@ -262,6 +263,14 @@ func (r *RunParameters) Restart() string {
 	return os.ExpandEnv(r.RawRestart)
 }
 
+func (r *RunParameters) SecurityOpt() []string {
+	var flags []string
+	for _, rawFlag := range r.RawSecurityOpt {
+		flags = append(flags, os.ExpandEnv(rawFlag))
+	}
+	return flags
```
michaelsauter: I prefer specific names. Maybe we can find a generic solution for these options which are handled all the same, but so far I haven't come up with one for Go. Until then I think it's better to use e.g. `securityOpt` here.

crane/container.go
```diff
@@ -536,6 +545,14 @@ func (c *container) createArgs() []string {
 	if c.RunParams.Rm {
 		args = append(args, "--rm")
 	}
+	// SecurityOpt
+	for _, flag := range c.RunParams.SecurityOpt() {
+		args = append(args, "--security-opt", flag)
+	}
```
michaelsauter: Same here.

michaelsauter:
Thanks for those!

bjaglin:
comment addressed in a separated commit

PR #147, opened 2015-02-03
add support for free-text extra args in run

bjaglin:
This can be useful when the crane flags mapping is lagging behind the latest Docker release. I figured out most of the happens within `docker run`, so that's where this option should be - I don't foresee much relevant activity in `rm` or `start`.


michaelsauter:
Hmm. My initial reaction is not to include this. I expect the pace of new `run` options to slow down. Also, I don't feel we save any time here since we would still need to do add support properly for each option.

I'm closing this for now, feel free to re-open if you disagree strongly :)

dreamcat4:
***+1.***

*Because Docker keep adding new run options on every new release. Docker 1.6 is no exception.*

PR #149, opened 2015-02-04
Support for optional dependencies and `required` cascading

bjaglin:
We have a need for optional dependencies as our microservices stack described by a `crane.yml` gets more and more complex. For example:
* A Crane user shouldn't have to start _all_ backends containers behind a proxy container
* Some stub containers are relevant only for testing but should remain within our main, centralized config.

Here is behavior of that new feature:
```
containers:
  foo:
    image: busybox
    run:
      detach: true
      cmd: ["sleep", "50"]
      link:
        - ?bar:bar
        - qux:qux
  bar:
    image: busybox
    run:
      detach: true
      cmd: ["sleep", "50"]
  qux:
    image: busybox
    run:
      detach: true
      cmd: ["sleep", "50"]
```

```
brice@t440s:~$ crane status foo -dall
NAME	IMAGE	ID	UP TO DATE	IP	PORTS	RUNNING
bar	busybox	-	-		-	-	-
qux	busybox	-	-		-	-	-
foo	busybox	-	-		-	-	-
```
```
brice@t440s:~$ crane status foo -drequired
NAME	IMAGE	ID	UP TO DATE	IP	PORTS	RUNNING
qux	busybox	-	-		-	-	-
foo	busybox	-	-		-	-	-
```
```
brice@t440s:~$ crane run -r -v foo
ERROR: Dependencies for container(s) foo could not be resolved. Check for cyclic or missing dependencies, or use -d/--cascade-dependencies to automatically attempt to recursively include dependencies in the set of targeted containers.
```
```
brice@t440s:~$ crane run -r -v foo -drequired
Command will be applied to: qux, foo

Running container qux ... 
--> docker run --detach --name qux busybox sleep 50
af4cd93da77fe57a29ec973c333b0d61f4ff6847e70d8d176df299f068b63321
Running container foo ... 
--> docker run --detach --link qux:qux --name foo busybox sleep 50
810d8f98cfc3477d10e8f56d5a463df8a421536432be50f895dd6f6cd4ee87de
```
```
brice@t440s:~$ crane run -r -v bar
Command will be applied to: bar

Running container bar ... 
--> docker run --detach --name bar busybox sleep 50
728e7e073b1d40077b2963f7f96a42edb8a2953ce8409b794ce2d4962949fd45
```
```
brice@t440s:~$ crane run -r -v foo
Command will be applied to: foo

Removing container foo ... 
--> docker rm --force foo
foo
Running container foo ... 
--> docker run --detach --link bar:bar --link qux:qux --name foo busybox sleep 50
e06b5fd7c1383d494cc5963d499ebe33c926cdd3f7edd654ab1251f7c79665e6
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,8 @@ If you want to use JSON instead of YAML, here's what a simple configuration look
 ```
 
 ## Advanced Usage
+
+### Groups
 Next to containers, you can also specify groups, and then execute Crane commands that only target those groups. If you do not specify any target as non-option arguments, the command will apply to all containers. However, you can override this by specifying a `default` group. Also, every container can be targeted individually by using the name of the container in the non-option arguments. Note that any number of group or container references can be used as target, and that ordering doesn't matter since containers will be ordered according to the dependency graph. Groups of containers can be specified like this (YAML shown):
 
 ```
@@ -188,7 +190,26 @@ groups:
 
 This could be used like so: `crane provision service1`, `crane run -v databases` or `crane lift -r services database1`. `crane status` is an alias for `crane status default`, which in that example is an alias for `crane status service1 database1`.
 
-When using targets, it is also possible to cascade the commands to related containers. There are 2 different flags, `--cascade-affected` and `--cascade-dependencies`. In our example configuration above, when targeting the `mysql` container, the `apache` container would be considered to be "affected". When targeting the `apache` container, the `mysql` container would be considered as a "dependency". Both flags take a string argument, which specifies which type of cascading is desired, options are `volumesFrom`, `link`, `net` and `all`.
+
+### Cascading
+When using targets, it is also possible to cascade the commands to related containers. There are 2 different flags, `--cascade-affected` and `--cascade-dependencies`. In our example configuration above, when targeting the `mysql` container, the `apache` container would be considered to be "affected". When targeting the `apache` container, the `mysql` container would be considered as a "dependency". Both flags take a string argument, which specifies which type of cascading is desired, options are `all`, `required`, `link`, `volumesFrom` and `net`.
+
+### Optional dependencies
+All dependencies are considered required, and thus enforced, unless their declaration is prefixed with a "?". Apart from allowing to get more control for selecting containers implicitly via cascading, when declaring a `link` (respectively a `volumesFrom`) optional, the dependency will be silently omitted if the target is not running (respectively not existing). In the following example, `foo` has an optional link dependency to `bar`, meaning that that link will only be created if `bar` already exists when starting `foo`, or if `bar` is part of the targeted containers list when starting `foo`, explictly or implicitly (via `--cascade-dependencies=all` for instance):
+```
+containers:
+  foo:
+    image: busybox
+    run:
+      detach: true
+      cmd: ["sleep", "50"]
+      link: ["?bar:bar"]
+  bar:
+    image: busybox
+    run:
+      detach: true
+      cmd: ["sleep", "50"]
+```
 
 ## Some Crane-backed sample environments
 * [Silex + Nginx/php-fpm + MySQL](https://github.com/michaelsauter/silex-crane-env)
--- a/crane/cmd.go
+++ b/crane/cmd.go
@@ -34,7 +34,7 @@ func isVerbose() bool {
 func configCommand(wrapped func(config Config), forceOrder bool) func(cmd *cobra.Command, args []string) {
 	return func(cmd *cobra.Command, args []string) {
 		for _, value := range []string{options.cascadeDependencies, options.cascadeAffected} {
-			if value != "none" && value != "all" && value != "link" && value != "volumesFrom" && value != "net" {
+			if value != "none" && value != "all" && value != "required" && value != "link" && value != "volumesFrom" && value != "net" {
 				print.Errorf("Error: invalid cascading value: %v", value)
 				cmd.Usage()
 				panic(StatusError{status: 64})
@@ -227,6 +227,7 @@ See the corresponding docker commands for more information.`,
 	craneCmd.PersistentFlags().StringVarP(&options.config, "config", "c", "", "Config file to read from")
 	cascadingValuesSuffix := `
 					"all": follow any kind of dependency
+					"required": like "all" without unresolved optional ones
 					"link": follow --link dependencies only
 					"volumesFrom": follow --volumesFrom dependencies only
 					"net": follow --net dependencies only
--- a/crane/container.go
+++ b/crane/container.go
@@ -130,31 +130,17 @@ func (c *container) Dependencies() *Dependencies {
 	dependencies := &Dependencies{}
 	for _, link := range c.RunParams.Link() {
 		linkParts = strings.Split(link, ":")
-		if !hasOptionalPrefix(&linkParts[0]) {
-			dependencies.Required = append(dependencies.Required, linkParts[0])
-		}
-		dependencies.All = append(dependencies.All, linkParts[0])
-		dependencies.Link = append(dependencies.Link, linkParts[0])
+		dependencies.add(linkParts[0], "link", !hasOptionalPrefix(&linkParts[0]))
 	}
 	for _, volumeFrom := range c.RunParams.VolumesFrom() {
-		if !dependencies.includes(volumeFrom) {
-			if !hasOptionalPrefix(&volumeFrom) {
-				dependencies.Required = append(dependencies.Required, volumeFrom)
-			}
-			dependencies.All = append(dependencies.All, volumeFrom)
-			dependencies.VolumesFrom = append(dependencies.VolumesFrom, volumeFrom)
-		}
+		dependencies.add(volumeFrom, "volumesFrom", !hasOptionalPrefix(&volumeFrom))
 	}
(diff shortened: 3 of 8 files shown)


michaelsauter:
I am sceptical about this one.

Mainly for 2 reasons:
1) This interferes with the 1:1 mapping of Crane options to Docker. If they would ever introduce something with a `?` in the links, we would have to change our behaviour to whatever theirs would be, breaking functionality for everyone who would be using this feature.
2) The dependency handling is already quite complex in my eyes. Introducing `required` would introduce even more complexity (both from a code and usage standpoint). Also, it is not orthogonal to the other types, since I can't say "lift all the required volumes-from containers" or similar. It's either `required` or `volumes-from`. The problem with this is that `required` isn't quite the same concept as `volumes-from`, `link` and `net`, I think.

That being said, I really appreciate your work and I would like to make Crane work for your use case if possible. Maybe you can give a real-life example on why you need this? Maybe then we can find another way to achieve the same goal.

bjaglin:
1) Good point there, although IMHO the chances are pretty low... I guess you believe introducing new fields `link-optional` and `volume-from-optional` wouldn't be any better?
2) Apart from a canonical mapping with docker commands, dependency handling is what differentiates Crane from other single-host orchestrating tools, so it does carry a bit of complexity. However, all this is opt-in and does not change the simple model. This is an advanced feature that most of the users wouldn't even run into, and I don't think the code is really more complex. Regarding the orthogonality type/required, would introduce another `--cascading-optional=true` flag address your concerns?

As for the uses cases I mentioned in the description, here is a more detailed explanation and some alternatives.

### A Crane user shouldn't have to start all backends containers behind a proxy container
Our central `crane.yml` is the single point of truth for the architecture of our product - its `crane graph` is the best tool we have to introduce new-comers to the team. Our public API is exposed via a haproxy mapping routes to a few underlying backends (service discovery being handled by `--links`), that are of course all up on production and on staging environments. However, it becomes increasingly memory-consuming to start all of them (`crane lift api -rdall`) on a local environment, where a developer rarely need more than one or two services behind the proxy. Currently the targets of all `links` declared in `crane.yml` need to be running for that container to be able to start.
* The current proposal allows a proxy to start with an arbitrary amount of underlying backends attached (the ones that have been manually started).
* Another alternative to achieve similar semantics would be to add a CLI flag `--ignore-missing-dependencies` (skipping `--volumes-from`, `--links` and `--net=container:<>` targeting respectively non-existing or non-running containers).

### Some stub containers are relevant only for testing but should remain within our main, centralized config
Another one of our component interacts with partners following a standardized API, which we need to stub for our integration tests. The idomatic way to have the IP of the stub be known by our component is to materialize a `link` in `crane.yml`. However, we do not want that stub to be running on local environments. Do we need to maintain 2 `crane.yml` in sync - one meant for staging with 25+ components, the other one for testing with these 25+ components and all stubs ?
* The current proposal allows to keep the stack within the same configuration, but gives the possibility to the user to leave out optional dependencies using `-drequired` 
* I experimented with a mix of environment variables magix & YAML `<<` merge keys (so that we would inject the optional `link`s only when wanted) but did not manage since `link` is an array which doesn't play well with merging.
* A new `--ignore-missing-dependencies` flag as suggested above could work, but doesn't play well with `-dall` for starting the top components and their direct/indirect dependency. However, we could combine it with a tagging mechanism (each container can be tagged with one, zero or several labels) and some `--cascade-ignore=<label>` used to avoid following `label=test` dependencies. I did not follow that route because I thought it was too generic/complex to solve the very specific use-case of optional dependency, but if you believe that's a better way to go, I'll be happy to work on a separate PR.

michaelsauter:
Thanks for the write-up. Very interesting!

First up, some comments:

> I guess you believe introducing new fields link-optional and volume-from-optional wouldn't be any better?

Correct, has it has the same problem, what if Docker would introduce those? And those sound more likely then the question mark :)

> Regarding the orthogonality type/required, would introduce another --cascading-optional=true flag address your concerns?

That would work I guess, but I imagine using it would feel very clumsy.

After reading through this, I wonder whether `required` would actually give you what you want. It sounds like some developers don't need the whole stack. But what exactly a dev needs might depend on what they are working on, so a "global" required setting won't do the trick - you might need more granular control.

Anyway, I like the idea of `--ignore-missing-dependencies` in general and was thinking a bit more about this. Also note that groups in Crane basically offer the same functionality as your idea of labels would.

How about adding an `exclude` flag which takes container or group references. These containers or groups are not started and (automatically or with an explicit --ignore-missing`) ignored if missing? E.g. `crane lift api --exclude testing` where testing is a group of containers not needed normally.

That would keep all this out of the Crane config and move it to the CLI, which I like. There might be a way to express this in the group definitions as well, but I'm afraid that's not possible without changing the format.

bjaglin:
Good point about reusing the group concept, I did have the same thinking _after_ I posted my suggestion above...

In the light of your analysis, I am now leaning towards a `--ignore-missing` and an extension of the target semantics: `foo bar ^qux` meaning including `foo`, `bar` but excluding `qux`(each of them being container or group references). That would workaround the lack of support for repeated argument in the current CLI lib (`--exclude` would be nice if only it could be repeated), and the evaluation of the include/exclude would happen in the order of declaration. We would still need to decide what `foo -dall ^qux` means in case qux is a dependency of `foo`, haven't given much though yet.

bjaglin:
Any thoughts around my last proposal? The last question remains central, as my use case calls for a way to remove containers when using cascading (and not statically defined groups): `foo -dall ^qux` would mean "foo, and all of its direct and indirect dependencies, skipping qux when traversing the graph".

michaelsauter:
Ha, I came to a different conclusion :) `--exclude` and `--ignore-missing` are not the same. Here's why:

Say there is group `a` with containers `x`, `y`, `z`, and group `b` with containers `x`, `y`. If you'd run `a ^b`, this would start `z`. All good so far. Now let's assume `z` depends on `y`, then this should fail, unless you explicitly ignore missing deps. 

I think `--ignore-missing` is a different concept then group/container exclusiion, even though it is related. So I would keep that separate. It is quite similar to the `forceOrder` stuff we already have, it just needs to be configurable from the CLI.

As to `--exclude` vs `^`, I kinda like the `^` as it is quite short and easy to understand.

bjaglin:
I didn't suggest that the concepts were the same, but merely that using exclusion would trigger a different default value for `--ignore-missing `. I guess that's because I view mostly (only?) a use case for exclusion together with cascading flags, in which case you do expect to have missing dependencies (see the scenarios that triggered my current implementation of optional dependencies). But anyway, I can live with an extra `-i/--ignore-missing` when using the feature.

michaelsauter:
@bjaglin Sorry, didn't mean to imply that you were. I too think automatic `--ignore-missing` is pretty much what you want when you use exclusion. But, if we made that the default, it would a) behave differently then the rest of the API, and b) we would need to find a way to opt-out. I don't want to do either of that, so I think the explicit opt-in `--ignore-missing` is the way to go.

bjaglin:
No problem at all! Regarding b), you could always do `--ignore-missing=false`, no? As for a) maybe I wasn't clear, but I was thinking about keeping `false` for the default, unless when one or several exclusion is used, in which case `--ignore-missing=true` is the default - is that this specific coupling you call "behaving differently"?

bjaglin:
One thing I am missing in the alternative currently discussed is that the crane user needs to have extensive knowledge of the graph if she wants to start only a subset of the dependencies for a given component.

Take the following graph:
```
- proxy
  - backend1
    - depA
    - depB
  - backend2
    - depC
    - depD
  - backend3
    - depE
    - depF
  - backend4
    - depG
    - depH
```
Ruling out the explicit materialization of all direct/indirect dependencies of a node of that tree as a group (which doesn't scale is error-prone and triggered the introduction of the cascading flag), how would you start the proxy and only a couple of its dependencies?

Apart from the obvious `crane run -i proxy backend1 depA depB backend2 depC depD`, we could do `crane run -idall proxy ^backend3 ^backend4`, but that implies that the user knows what shouldn't be started, either explicitly or via a group (mapping my second scenario) - as opposed as to what should be started (mapping to my first scenario) as `crane run -idall proxy backend1 backend2` wouldn't skip the unwanted backends. In other words, combining whitelists and cascading seems impossible with the current solution.

michaelsauter:
I see your point. Very tricky indeed :)

However, I'm not quite sure I understand why you're opposed to the idea that the user needs to know what should not be started? To me, this whole feature sounds like an explicit opt-in optimisation (by choosing to ignore parts of the graph). This will potentially break things, therefore the user should be aware of it. As far as I understand it, your specific use case has to do with "testing" backends, so putting all of those into a group doesn't sound to complicated either.

Your suggested alternative would be quite powerful indeed, but I agree that is looks very complex.

bjaglin:
So after thinking it through, I guess we can just live with 2 separate commands for whitelisting (as opposed to blacklisting via `^`):
```
crane run -dall backend1 backend2
crane run -i proxy
```

I'll move forward and implement a string flag `-i/--ignore-missing` (taking the same values as `--cascade-*`) and the exclusion prefix in 2 different PRs. Keeping this open until the use cases presented here are supported.

bjaglin:
this will be obsoleted by https://github.com/michaelsauter/crane/issues/207

PR #151, opened 2015-02-10
Fix sig-proxy mapping

bjaglin:
The mapping for `--sig-proxy` introduced in https://github.com/michaelsauter/crane/commit/2827f11a610dc1c43f23ec61afe7f5ea2fa958d3 and available in crane 1.1.0 is broken, as the value, whether it was unset, set to `true` or set to `false`, had the effect of using the default (`true`) value in Docker :(

1c70909 fixes the obviously-wrong behavior of `false`, but introduces an issue when the field is unsed (defaulting to `false`)... Therefore this is not in any way a final solution to the problem. I did not manage to find a way to have default `true` value for missing `bool` keys from json/YAML. Any idea how we can map that nicely without having to break the one-one mapping and introduce a NoSigProxy in `crane.yml`? AFAIK, `--sig-proxy` is the only boolean docker flag with `true` as default...

### DIFF
--- a/crane/container.go
+++ b/crane/container.go
@@ -1,6 +1,7 @@
 package crane
 
 import (
+	"encoding/json"
 	"fmt"
 	"github.com/flynn/go-shlex"
 	"github.com/michaelsauter/crane/print"
@@ -77,7 +78,7 @@ type RunParameters struct {
 	RawRestart     string      `json:"restart" yaml:"restart"`
 	Rm             bool        `json:"rm" yaml:"rm"`
 	RawSecurityOpt []string    `json:"security-opt" yaml:"security-opt"`
-	SigProxy       bool        `json:"sig-proxy" yaml:"sig-proxy"`
+	SigProxy       OptBool     `json:"sig-proxy" yaml:"sig-proxy"`
 	Tty            bool        `json:"tty" yaml:"tty"`
 	RawUser        string      `json:"user" yaml:"user"`
 	RawVolume      []string    `json:"volume" yaml:"volume"`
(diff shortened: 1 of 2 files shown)


michaelsauter:
Oh right. I don't know of a way around this as well. I think I've run into this before ... Go is pretty limited there. Would be great to have the ability to specify a default.

Anyway, I think your fix introduces a bigger problem. Right now, that field is broken, but with this fix the behaviour of all containers managed by Crane would change. 

One way to do this might be to use a string instead of a boolean? Then it would become
```
if c.RunParams.SigProxy == "false" {
    args = append(args, "--sig-proxy=false")
}
```

Horrible, but maybe acceptable? YAML users wouldn't notice the difference ... :)

bjaglin:
That seems like a good compromise. I'll do that if I can't find anything better - I am hoping a custom, three-way struct (unset, set to true, or set to false) with custom YAML and JSON unmarshaller could still do the trick and do the job without compromising on the config type.

michaelsauter:
Phew, sounds complex :) I'd probably just do the string ...

bjaglin:
@michaelsauter that worked! :)

michaelsauter:
Oh that's nice! Great work.

Happy to merge, but wanted to ask whether you would want to extract that into a generic package like the multiplex stuff? It's super small, but might be useful for others too.

bjaglin:
IMHO, it's really too small to extract to a separate git repo, and I don't see how to make it "generic"... However, I guess the struct could go to a separate file/folder (so that imports look like `github.com/michaelsauter/crane/unmarshalling` for example)... What do you think?

michaelsauter:
Ah okay. No I think in this case we leave it as is for now.

PR #152, opened 2015-02-11
Add the cmd flag to the Run and Create commands

sgirones:
Cmd flag will overwrite the command defined on the Crane configuration

The use case is a Continuous Integration job that runs commands depending on the environment. It's not practical to create an specific container for each combination.
This way you can define an common environment for many commands.

Example:
```
containers:
  test:
    image: busybox
    run:
      cmd: ["echo", "foo"]
      rm: true
```
```
$ crane run test
Running container test ... foo
$ crane run test --cmd="echo bar"
Running container test ... bar
```

### DIFF
--- a/crane/cmd.go
+++ b/crane/cmd.go
@@ -22,6 +22,7 @@ type Options struct {
 	cascadeAffected     string
 	config              string
 	target              []string
+	cmd                 string
 }
 
 var options Options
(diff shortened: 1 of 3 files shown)


bjaglin:
We have a somewhat similar use-case, and we are using this, which works fine:
```
containers:
  test:
    image: busybox
    run:
      cmd: $TEST_CMD
      rm: true

```

```
$ TEST_CMD="echo foo" crane  run
Running container test ... foo
```

IMHO, introducing a flag that override  the `cmd` attribute of ALL targeted containers does not really map with the current semantics.

sgirones:
I tried to find a better way to do that, didn't try env variables... thanks @bjaglin 

Closing PR.

bjaglin:
thanks for the contribution though!

PR #153, opened 2015-02-13
Update README.md

dreamcat4:
Build instructions for users building crane from master branch.


michaelsauter:
Thanks for this. Could definitely be a bit better. However, I don't think these instructions are spot on nor should we impose that much. I would prefer to link to relevant Go docs and let the user decide how they want to set up their environment.

README.md
```diff
@@ -5,15 +5,32 @@ Lift containers with ease
 Crane is a tool to orchestrate Docker containers. It works by reading in some configuration (JSON or YAML) which describes how to obtain images and how to run containers. This simplifies setting up a development environment a lot as you don't have to bring up every container manually, remembering all the arguments you need to pass. By storing the configuration next to the data and the app(s) in a repository, you can easily share the whole environment.
 
 ## Installation
-The latest release can be installed via:
+The latest releases bre-built in binary form can be installed via:
```
mishak87: bre :arrow_right: pre

michaelsauter:
Closing this as it is not amerceable-as is. Any further discussion can happen in #167.

PR #154, opened 2015-02-21
Add new flag: -i/--ignore-missing={none|all|link|volumesFrom|net}

bjaglin:
Related discussion in https://github.com/michaelsauter/crane/pull/149

### DIFF
--- a/crane/cmd.go
+++ b/crane/cmd.go
@@ -20,6 +20,7 @@ type Options struct {
 	colorize            bool
 	cascadeDependencies string
 	cascadeAffected     string
+	ignoreMissing       string
 	config              string
 	target              []string
 }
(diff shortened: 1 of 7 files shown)


michaelsauter:
Thanks for your work.

Looking at the code and the corner cases, I am unsure about it again. Feels to complicated for me.

I will have to think a little more about this ...

bjaglin:
I could only find one corner case, as mentioned in the comments (several dependencies of different kinds from one container to another).

I think most (all?) of the complexity comes from the `order` method, which could be greatly simplified, if we simply rely on the actual run to detect missing dependencies (which would be more verbose since it gives more context as to exactly what is missing where). I wanted to have a shot at that, but I thought this was outside the scope of this PR. Want me to do it here?

We badly need this feature as I explained in the original PR, and it needs to be part of core - I thought about doing some YAML preprocessing in a separate tool, but we would need to check for container existence and traverse the graph anyway, so if we go that way, we might just as well fork. I am willing to rework another time if you think an approach would match more the vision/roadmap.

bjaglin:
rebased

michaelsauter:
Thanks. I just had another look and it looks much better now. Wondering why I was so hesitant in the first place.

The only thing I'm unsure is whether we overcomplicate this by allowing the user to select what to ignore. Wouldn't a simple true/false be enough?

Once that's settled, happy to merge and include it in a (hopefully not far away) next release.

bjaglin:
That was my original idea (see https://github.com/michaelsauter/crane/pull/149), but when I started the implementation, I realized that this feature was related to the cascading flags since they both define a way to traverse the dependency graph, so I though it would be better to align the values the flags can take. As you might recall, I did not have any strong use-case for that level of granularity at the time we introduced the cascading flags, and neither do I have for this, but I feel they go together.

bjaglin:
for example: `crane lift -r -dlink -ivolumesFrom` or `crane lift -r -ivolumesFrom` to lift a container safely (i.e. with its dependencies) but without the backup/bootstrap data it gets via volumesFrom?

michaelsauter:
OK, let's leave it like it is for now. 

Sorry for the long wait, and thanks again for this!

Issue #74, opened 2014-07-24
Opt-in feature: block until exposed ports are bound when starting containers


bjaglin:
Proposal
-----------
Allow new container flags `hooks.run-start.pre` &  `hooks.run-start.post` in the config, which are (potentially blocking) shell commands to be executed in the foreground before and after the `docker run` or `docker start`  command for that container. This allow for custom waiting/presence/sidekicking code. Return code from these commands is checked so they can interrupt the flow.

Use-case
-------------
For example, if I want `foo` to be started only if/after the port `1234` of `bar` is bound:
```
containers:
  foo
    image: foo
    run:
      tty: true
      interactive: true
      link:
        - bar:bar
    hooks:
      run-start:
        pre: "docker run -ti --rm --link bar:bar busybox sh -c 'for i in `seq 1 10`; do sleep 1 && nc bar 1234 && exit 0 || echo retrying...; done; exit 1'"

  bar
    image: bar
    run:
      expose:
        - 1234
```

To accomplish this currently, I have to define an extra container in between the producer/server `bar` and consumer/client `foo`, and define a dummy `run.volumse-from` from the consumer/client to that middle container to explicitly declare the dependency.

```
containers:
  foo
    image: foo
    run:
      tty: true
      interactive: true
      link:
        - bar:bar
      volumes-from:
        - wait-for-bar

  wait-for-bar
    image: busybox
    run:
      tty: true
      interactive: true
      cmd: ["sh", "-c", "for i in `seq 1 10`; do sleep 1 && nc bar 1234 && exit 0 || echo retrying...; done; exit 1"]
      link:
        - bar:bar

  bar
    image: bar
    run:
      expose:
        - 1234
```

I think this is a common use case for single-host orchestration, and this is generic enough to cover many use cases.

bjaglin:
Let's follow up on https://github.com/michaelsauter/crane/issues/148

PR #159, opened 2015-03-02
add support for container hooks, defined on groups or containers

### LINKED ISSUE #148: Host run hooks for pre-cmd, post-cmd etc.
It would be great to have some way to add host-side commands to run before `crane lift` `crane provision` etc. And post hooks too.

This allows us to continue using Crane CLI interactively during development, without wrapping the `crane` cmd in some other custom, archaic or cumbersome script.

There are many uses. Pretty much anything else that can't be run in a container, which must be run on the host side.

For example:

* To run pipework and add network interfaces to containers when they are started.
* To run a command to check that a service has come up successfully after crane lift.
* To do something (like push or publish new images, tweet them), after crane provision.

Guess it could be another key in the `crane.yml` or something. Haven't thought about it very much. 

For one thing, it would be necessary to decide whether it is worthwhile to differentiate between global, per-group, and per-container pre/post hooks. And which specific set of actions (provision, start, etc) need to have hooks.

bjaglin:
Fixes #148. As mentioned in another open PR, I have a couple of ideas for refactoring the code which is starting to smell in `config.go`, but I tried to keep this PR straight to the feature scope, to handle the refactoring separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,9 @@ If you want to use JSON instead of YAML, here's what a simple configuration look
 ```
 
 ## Advanced Usage
+
+### Groups
+
 Next to containers, you can also specify groups, and then execute Crane commands that only target those groups. If you do not specify any target as non-option arguments, the command will apply to all containers. However, you can override this by specifying a `default` group. Also, every container can be targeted individually by using the name of the container in the non-option arguments. Note that any number of group or container references can be used as target, and that ordering doesn't matter since containers will be ordered according to the dependency graph. Groups of containers can be specified like this (YAML shown):
 
 ```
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -188,8 +191,59 @@ groups:
...
+hooks:
+  foo:
+    post-start: echo container from foo started
+  bar:
+    post-stop: echo container from bar stopped
+  service3:
+    post-stop: echo container service3 stopped
+```
+
+Hooks can be defined on a group level (`foo`, `bar`) so that they apply to all containers within that group, or directly on a container (`service3`). At most one hook can be registered per container and per event. When more than one hook is found for a given container and a given event, the following rules apply:
```
bjaglin: I restricted the usage of several hooks compared to what was discussed in #148, to keep implementation concise and avoid non-deterministic behavior.

michaelsauter:
Thanks for this - it will probably take a little while until I can have a look. Sorry for the delay. Looking forward to get this in though!

bjaglin:
Thanks for the heads up! Any rough ETA - are we talking about days or weeks before you get the chance to have a look?

michaelsauter:
Probably weeks - I'm in the middle of moving overseas ...

adrianhurt:
Hi, I'm new in this dockerized world. There are loads of options to composing containers, and finally I found Crane yesterday. I think it's awesome. I really appreciate your work. And I found it googling a lot trying to find this exact problem. I think it is a very useful feature to be able to execute some code or scripts before or after starting or stoping containers, especially to checking if some process is ready or logging.

Is there any possibility to try an executable for this PR? I know it's not yet "official" but I would like to try it. Or maybe some instruction to be able to compile it from the code… I don't know Go at all.

Thanks to both!

bjaglin:
@adrianhurt maybe https://github.com/michaelsauter/crane/pull/153/files can help you to get started and build this locally?

adrianhurt:
Thanks for the quick response! But I get a problem following these steps. I've tried within an empty ubuntu:14.04 container installing git and following these steps and I get while executing `go get`:

```
unexpected fault address 0x7f7500000011
fatal error: fault
[signal 0xb code=0x1 addr=0x7f7500000011]

goroutine 5 [running]:
	:0
	:0
...
	:0

goroutine 1 [select]:
main.$nested27
	/build/buildd/gccgo-go-1.2.1/src/cmd/go/http.go:57
main.httpsOrHTTP
	/build/buildd/gccgo-go-1.2.1/src/cmd/go/http.go:65
main.repoRootForImportDynamic
	/build/buildd/gccgo-go-1.2.1/src/cmd/go/vcs.go:468
main.repoRootForImportPath
	/build/buildd/gccgo-go-1.2.1/src/cmd/go/vcs.go:362
main.downloadPackage
	/build/buildd/gccgo-go-1.2.1/src/cmd/go/get.go:269
main.download
	/build/buildd/gccgo-go-1.2.1/src/cmd/go/get.go:166
main.download
	/build/buildd/gccgo-go-1.2.1/src/cmd/go/get.go:232
main.download
	/build/buildd/gccgo-go-1.2.1/src/cmd/go/get.go:232
main.runGet
	/build/buildd/gccgo-go-1.2.1/src/cmd/go/get.go:72
main.main
	/build/buildd/gccgo-go-1.2.1/src/cmd/go/main.go:161

goroutine 3 [syscall]:
	goroutine in C code; stack unavailable
```

bjaglin:
If you are having troubles building, you can just take the verbose approach for now, which is to define an interactive container with a fake dependency to the one you want to add a hook on. That's what I am living with until this gets accepted, see https://github.com/michaelsauter/crane/issues/74.

adrianhurt:
I'll do that meanwhile I wait for this feature. Thanks!

michaelsauter:
Sorry for the long wait. I had a quick look and think this is great. 

Thanks @bjaglin!

PR #164, opened 2015-03-20
Add ps command as alias of status

mattheath:
Other tools such as _docker_ and _fig_ use the `ps` command to list the status of containers, rather than the `status` command that crane uses. This adds a `ps` command with the same functionality as the `status` command, so that my muscle memory works better :wink: 

Couldn't see an easy way to alias commands using _cobra_, so I've added `ps` as a copy of `status`.

### DIFF
--- a/crane/cmd.go
+++ b/crane/cmd.go
@@ -183,6 +183,15 @@ formatted in bold provided the 'colorize' flag is on.`,
 		}, true),
 	}
 
+	var cmdPs = &cobra.Command{
+		Use:   "ps",
+		Short: "Displays status of containers",
+		Long:  `Displays the current status of all targeted containers.`,
+		Run: configCommand(func(config Config) {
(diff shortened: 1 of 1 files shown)


adrianhurt:
Hi @mattheath. In my opinion, the problem is that `crane status` gives the information of every container with a different output format than `docker ps`. Then I would expect that a `crane ps` functionality would give the same format as `docker ps` but only for those containers that are defined within your `crane.yml` file.

bjaglin:
I tend to agree with @adrianhurt - if a `crane` command matches a `docker` command, it has to be a simple wrapper. If I had to guess `crane ps [<target1> [<target2> ...]]` would just show the `docker ps` limited to the targets, so I am rather against making that alias.

michaelsauter:
I agree with @adrianhurt and @bjaglin. Crane should always map 1:1. I'm not against a new `ps` command however which maps to `docker ps`.

PR #165, opened 2015-03-24
Update README.md to include YAML advanced features

adrianhurt:


README.md
```diff
@@ -192,6 +192,25 @@ This could be used like so: `crane provision service1`, `crane run -v databases`
...
+      link: ["db:db"]
+      ...
+      cmd: web
+  web2: *web
+
+  admin1: &admin { <<: *web, run: { <<: *web-run , cmd: admin }}
+  admin2: *admin
+```
+
+As a summary, `&anchor` declares the anchor property, `*alias` is the alias indicator to simply copy the mapping it references, and `<<: *merge` includes all the mapping but letting you to override some keys.
```
bjaglin: "but let you to override some keys"

README.md
```diff
@@ -192,6 +192,25 @@ This could be used like so: `crane provision service1`, `crane run -v databases`
 
 When using targets, it is also possible to cascade the commands to related containers. There are 2 different flags, `--cascade-affected` and `--cascade-dependencies`. In our example configuration above, when targeting the `mysql` container, the `apache` container would be considered to be "affected". When targeting the `apache` container, the `mysql` container would be considered as a "dependency". Both flags take a string argument, which specifies which type of cascading is desired, options are `volumesFrom`, `link`, `net` and `all`.
 
+### YAML advanced usage
+YAML gives you some advanced features like [alias](http://www.yaml.org/spec/1.2/spec.html#id2786196) and [merge](http://yaml.org/type/merge.html). They lets you a way to easily avoid duplicated code in your `crane.yml` file. As a example, imagine you need to define 2 different webapps: `web` and `admin`. They share almost the same configuration but the `cmd` declaration. And imagine you also need 2 instances for each one for using with a node balancer. Then you can declare them as simply:
```
bjaglin: "They allow you to easily avoid duplicated code"

bjaglin: "2 different webapps" -> "2 different containers"

bjaglin: "And imagine you also need 2 instances for each one for using with a node balancer. Then you can declare them as simply:" -> "In case you need several instances of the same image to use them behind a load balancer for example, you can simply declare them as:"

bjaglin:
Thanks! I am just a bit hesitant since the feature described (namely override a nested value from a merged source directly in the target) works great in crane, but relies on a faulty implementation of the YAML spec in the version of `go-yaml` currently linked... See https://github.com/go-yaml/yaml/issues/81 for details.

adrianhurt:
Thanks for the grammar corrections. They should be fine now. My English is not as good as I would like.

About the faulty implementation you say… I don't understand very well what are you meaning in this case. In the example I have 2 different aliases (`&web` and `&web-run`) in order to merge correctly the nested values. I think it is the correct behavior described in [merge specification](http://yaml.org/type/merge.html). But as I understand seeing your [go-yaml/yaml#81](https://github.com/go-yaml/yaml/issues/81) issue, the problem would appear if the merge is done after declaring the overriding values. Maybe we could warn it as a note...

README.md
```diff
@@ -192,6 +192,25 @@ This could be used like so: `crane provision service1`, `crane run -v databases`
...
+      link: ["db:db"]
+      ...
+      cmd: web
+  web2: *web
+
+  admin1: &admin { <<: *web, run: { <<: *web-run , cmd: admin }}
+  admin2: *admin
+```
+
+As a summary, `&anchor` declares the anchor property, `*alias` is the alias indicator to simply copy the mapping it references, and `<<: *merge` includes all the mapping but let you to override some keys.
```
bjaglin: let you override some keys

bjaglin:
Right, my bad, I didn't realize you had 2 different anchors, and therefore nothing nested in the anchor values. Thanks for the update!

michaelsauter:
Thanks @adrianhurt for this! And sorry for the long wait ...

Can you rebase on master so I can merge this? It would be ideal if you could squash the second commit in the first one too.

adrianhurt:
Hi @michaelsauter . I think it's now ready for merging

Issue #163, opened 2015-03-19
Define variables within crane.yml


adrianhurt:
I know you can use variables within a `crane.yml` directly as it was a shell script using `$` character. But it implies you have to declare them previously apart from the file. I think it could be interesting to have the chance to declare them also directly within the file. Then the deployment would be more portable. An example could be:
```
variables:
  - db_user=root
  - db_password=rootpass
  - db_database=my_database

containers:
  mysql:
    run:
      env:
        - MYSQL_ROOT_PASSWORD=@db_password
        - MYSQL_DATABASE= @db_database
  mysql-backup:
    run:
      link: "mysql:mysql"
      volume: ["/mysql/backups:/mysql/backups"]
      env:
        - MYSQL_USER=@db_user
        - MYSQL_PASSWORD=@db_password
        - MYSQL_DATABASE=@db_database
  web:
    run:
      link: ["mysql:mysql"]
      env:
        - MYSQL_USER=@db_user
        - MYSQL_PASSWORD=@db_password
        - MYSQL_DATABASE=@db_database
```
Another possibility could be using directly the `$` character and parse each variable and export it directly within the scope.

I really would like to help more and uploading the corresponding PR, but I don't have enough time now to start with Go language…

Thanks again for the project!

bjaglin:
You can do something very similar just by using a YAML feature: http://yaml.org/type/merge.html

adrianhurt:
Cool!, I didn't know the alias and merging features of YAML. As I see you can do this:
```
aliases:
  - &db_user MYSQL_USER=root
  - &db_password MYSQL_PASSWORD=rootpass
  - &db_database MYSQL_DATABASE=my_database

containers:
  web:
    run:
      link: ["mysql:mysql"]
      env:
        - *db_user
        - *db_password
        - *db_database
```

However it implies that every key in the corresponding `key=value` string should be the same. I was trying to make more with this features but it seems the fact of using a string array for `env` stops us. I know it's a direct mapping to `-e` parameter of Docker, but if we could also declare the `env` as a mapping (and covert it finally into a `key=value` string, we could do some interesting things as the following examples.
```
aliases:
  - &db_user root
  - &db_password rootpass
  - &db_database my_database

containers:
  mysql:
    run:
      env:
        MYSQL_ROOT_PASSWORD: *db_password
        MYSQL_DATABASE: *db_database
  web:
    run:
      link: ["mysql:mysql"]
      env:
        MYSQL_USER: *db_user
[...]

PR #168, opened 2015-03-25
Adding support for env as a mapping

adrianhurt:
It allows to declare the `env` as a mapping in addition to the array syntax. It lets declare alias in the yaml file and use it as variables and also merge blocks easily.

See more and examples in issue #163 (https://github.com/michaelsauter/crane/issues/163)

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,9 @@ The map of containers consists of the name of the container mapped to the contai
 	* `attach` (boolean)
 	* `interactive` (boolean)
 
-See the [Docker documentation](http://docs.docker.io/en/latest/reference/commandline/cli/#run) for more details about the parameters.
+Note that basic environment variable expansion (`${FOO}`, `$FOO`) is supported throughout the configuration, but advanced shell features such as command substitution (`$(cat foo)`, `` `cat foo` ``) or advanced expansions (`sp{el,il,al}l`, `foo*`, `~/project`, `$((A * B))`, `${PARAMETER#PATTERN}`) are *not* as the Docker CLI is called directly.
+
+See the [Docker documentation](http://docs.docker.io/en/latest/reference/commandline/cli/#run) for more details about the parameters. 
 
 ## Example
 A typical `crane.yaml` looks like this:


adrianhurt:
I would upload another PR to update `README.md` in case it is merged

crane/container.go
```diff
@@ -206,8 +206,19 @@ func (r *RunParameters) Entrypoint() string {
 
 func (r *RunParameters) Env() []string {
 	var env []string
-	for _, rawEnv := range r.RawEnv {
-		env = append(env, os.ExpandEnv(rawEnv))
+	if r.RawEnv != nil {
+		switch rawEnv := r.RawEnv.(type) {
+  	  case []interface{}:
```
bjaglin: please run `gofmt` before committing

crane/container.go
```diff
@@ -206,8 +206,19 @@ func (r *RunParameters) Entrypoint() string {
...
+  			for _, v := range rawEnv {
+  				env = append(env, os.ExpandEnv(v.(string)))
+  			}
+      case map[interface{}]interface{}:
+  			for k, v := range rawEnv {
+          env = append(env, os.ExpandEnv(k.(string)) + "=" + os.ExpandEnv(v.(string)))
```
bjaglin: maybe this needs a bit more error handling in case the casting fails?

crane/container.go
```diff
@@ -206,8 +206,19 @@ func (r *RunParameters) Entrypoint() string {
...
+  			}
+  		default:
+  			print.Errorf("env is of unknown type!")
```
bjaglin: that should be a fatal error (panic) and the faulty value should be passed along

bjaglin:
I like the idea! Could you maybe update existing unit tests passing JSON/yaml configs to document/test that feature against regression?

adrianhurt:
I've just added a TestEnv to test the array and mapping options. I've also run `gofmt`. However, as you note, I don't know almost nothing about Go. This is the first time I use it. So I've simply copied, pasted and readapted the code I've seen for `cmd` as you suggested me. So the error handling and test are very similar to that ones. I've only adapted it to this feature.

README.md
```diff
@@ -81,7 +81,7 @@ The map of containers consists of the name of the container mapped to the contai
 	* `device` (array) Add host devices.
 	* `dns` (array)
 	* `entrypoint` (string)
-	* `env` (array)
+	* `env` (array/mapping) `env` can be declared as a mapping where each `key`-`value` will be translated to a string `"key=value"`.
```
bjaglin: I think it should be made more clear that the type of `env` is _either_ an array of strings or a mapping of strings - and that the example given should be declined in both versions since none is "preferred".

adrianhurt: Maybe with something like:
* `env` (array/mapping) It can be declared as a string array with `"key=value"` format or a mapping where each `key: value` will be translated to the corresponding `"key=value"` string.

And for the example, maybe we could add a note below indicating that env could be also declared as:
```
env:
  MYSQL_ROOT_PASSWORD: mysecretpassword
```
I don't think it would be necessary to copy again the whole example.

In addition, if this PR is merged, the other one for the advanced features of YAML could be updated to show another example using this mapping-env with anchors behaving as variables.

michaelsauter:
Thanks, I think I'm generally in favour of this but need to have a proper look. Will be another couple of days/weeks. Sorry about that, but I will get to it.

adrianhurt:
Of course, don't worry!

README.md
michaelsauter: @adrianhurt I like your latest suggestion. If you update the text I'll merge this in.

PR #169, opened 2015-03-25
Shortcut to allow to define a link as "name" instead of "name:alias" when name and alias are the same

adrianhurt:
Example:
```
link: ["mysql", "elasticsearch"]
```
instead of
```
link: ["mysql:mysql", "elasticsearch:elasticsearch"]
```

### DIFF
--- a/crane/container.go
+++ b/crane/container.go
@@ -235,6 +235,9 @@ func (r *RunParameters) Hostname() string {
 func (r *RunParameters) Link() []string {
 	var link []string
 	for _, rawLink := range r.RawLink {
+    if len(strings.Split(rawLink, ":")) == 1 {
+    	rawLink += ":" + rawLink
+    }
 		link = append(link, os.ExpandEnv(rawLink))
 	}
 	return link


bjaglin:
Not convinced about that one - what if future releases of docker introduce a new valid value for `link` that does not include `:` and has different semantics than the one you suggest to implement on the crane side? I think the shortcut is not worth the risk.

adrianhurt:
I see your point… I close this then. Thanks!

bjaglin:
Thanks though for taking the time!

adrianhurt:
Haha, it was only a 2-line PR… However, I've just seen it's already supported in Docker Compose ([here](https://docs.docker.com/compose/yml/#links)). In deed, I've just made a new [proposal](https://github.com/docker/docker/issues/11824) directly to Docker, hehe. Let's see what they say...

adrianhurt:
Finally, the proposal was done by Docker's team. So we'll be able to use it soon without change anything in crane

bjaglin:
Great to get it upstream! Just added https://github.com/michaelsauter/crane/pull/190 to ensure compatibility.

Issue #155, opened 2015-02-24
dockerfile: ability to specify multiple Dockerfiles in the same (sub)directory


dreamcat4:
Have been talking on `#docker` IRC about this today. Some people need to have 'variant' images. Where the Dockerfile is nearly the same (tweaked a bit). So then it's neater to share the context (support files) and have multiple Dockerfiles in the same directory.

The syntax needed looks like this:

```sh
docker build -f dir/Dockerfile.2 dir

# where dockerfile: dir/Dockerfile.2
```

rather than

```sh
docker build dir

# where dockerfile: dir
```

Reading from `man docker-build` this can be done with the existing `dockerfile:` configuration key if testing for file or directory. If `dockerfile:` resolves to being a directory (e.g. `test -d`) then use the 2nd command. As crane currently does always.

If `dockerfile:` resolves to being a file (e.g. `test -f`). Then get the directors components (everything up until the last `/`). Or use `.` for `dir` if there are no slashes in the value of `dockerfile:`. Which is a corner case. Then use the first above command with `-f` flag ^^.

Sorry if that explanation isn't terribly clear. Here is the place where the source code is running the `docker build` command:

https://github.com/michaelsauter/crane/blob/master/crane/container.go#L732-740

michaelsauter:
I think I got it :)

Makes sense to me, and would be possible to introduce without BC breaks, so I'm :+1: 

I'm in the middle of moving overseas, so no time at all to implement this right now. Do you want to give it a shot?

dreamcat4:
Don't quite have the time either ATM. But sure - if (more likely when) I end up needing it in the future. Unless someone else gets there first.

dreamcat4:
Right. Well it seems real life is not so simple as we might hope.

Due to URLs being also a valid build source. After another look at this. And going inside the actual source code of the `docker build` command and docker API. We cannot fully support docker build options (the command line OR build/ api) in the originally proposed way.

@michaelsauter So I will need you to decide upon a new variable name for the `PATH | URL` mandatory build argument. Remembering that it sets the docker build context. You have 2 best choose from: Either `context:` or `location:`. Please choose one of them.

This new approach also helps us to properlay support any future functionality that may get added to the `docker build` command.

[...]

PR #172, opened 2015-03-28
New location: key, smart handling of 'location:' & 'dockerfile:' #155

dreamcat4:
Ok. Here is my pull request. I have tested this with `crane provision -v` to see the arguments being sent to `docker build` command. My test run:

```sh
id@emachines-e520:~/docker-images$ crane provision -v context
Command will be applied to: context

Building image dreamcat4/context ... 
--> docker build --rm --tag=dreamcat4/context --file=context/context/Dockerfile context
Sending build context to Docker daemon 4.608 kB
Sending build context to Docker daemon 
Step 0 : FROM busybox
 ---> 4986bf8c1536
Step 1 : RUN echo "context/context/Dockerfile (subdirectory)"
 ---> Using cache
 ---> 92809b6040b1
Successfully built 92809b6040b1
id@emachines-e520:~/docker-images$ crane provision -v context
Command will be applied to: context

Building image dreamcat4/context ... 
--> docker build --rm --tag=dreamcat4/context --file=context/Dockerfile context
Sending build context to Docker daemon 4.608 kB
Sending build context to Docker daemon 
Step 0 : FROM busybox
 ---> 4986bf8c1536
Step 1 : RUN echo "context/Dockerfile (parent folder)"
 ---> Using cache
 ---> 251170debed2
Successfully built 251170debed2
id@emachines-e520:~/docker-images$ crane provision -v context
Command will be applied to: context

Building image dreamcat4/context ... 
--> docker build --rm --tag=dreamcat4/context --file=context/Dockerfile.BBB context
Sending build context to Docker daemon 4.608 kB
Sending build context to Docker daemon 
Step 0 : FROM busybox
 ---> 4986bf8c1536
Step 1 : RUN echo "context/Dockerfile.BBB (parent folder)"
 ---> Running in 85eb3d87b1e4
context/Dockerfile.BBB (parent folder)
 ---> 58d909d64a69
Removing intermediate container 85eb3d87b1e4
Successfully built 58d909d64a69
id@emachines-e520:~/docker-images$ crane provision -v context
Command will be applied to: context

Building image dreamcat4/context ... 
--> docker build --rm --tag=dreamcat4/context --file=context/Dockerfile.BBB context
Sending build context to Docker daemon 4.608 kB
Sending build context to Docker daemon 
Step 0 : FROM busybox
 ---> 4986bf8c1536
Step 1 : RUN echo "context/Dockerfile.BBB (parent folder)"
 ---> Using cache
 ---> 58d909d64a69
Successfully built 58d909d64a69
id@emachines-e520:~/docker-images$ crane provision -v context
Command will be applied to: context

Building image dreamcat4/context ... 
--> docker build --rm --tag=dreamcat4/context --file=context/Dockerfile context
Sending build context to Docker daemon 4.608 kB
Sending build context to Docker daemon 
Step 0 : FROM busybox
 ---> 4986bf8c1536
Step 1 : RUN echo "context/Dockerfile (parent folder)"
 ---> Using cache
 ---> 251170debed2
Successfully built 251170debed2
id@emachines-e520:~/docker-images$ crane provision -v context
Command will be applied to: context

Building image dreamcat4/context ... 
--> docker build --rm --tag=dreamcat4/context context
Sending build context to Docker daemon 4.608 kB
Sending build context to Docker daemon 
Step 0 : FROM busybox
 ---> 4986bf8c1536
Step 1 : RUN echo "context/Dockerfile (parent folder)"
 ---> Using cache
 ---> 251170debed2
Successfully built 251170debed2
id@emachines-e520:~/docker-images$ 
```

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -10,3 +10,4 @@ Mathew Davies <thepixeldeveloper@googlemail.com>
 Scott M. Likens <scott@likens.us>
 eggtree <eggtree@requiredhealth.org>
 gissehel <public-devgit-dantus@gissehel.org>
+Dreamcat4 <dreamcat4@gmail.com>
--- a/README.md
+++ b/README.md
@@ -42,7 +42,11 @@ Maps to `docker pause`.
 Maps to `docker unpause`.
 
 ### `provision`
-Either calls Docker's `build` or `pull`, depending on whether a Dockerfile is specified. The Docker cache can be disabled by passing `--no-cache`.
+Calls Docker's `build` command if a `location:` or `dockerfile:` was specified. Else is there is only an image name to go on, calls Docker's `pull` command.
+
+***Note regarding backwards compatibility:***
+
+Previously only the `dockerfile:` key was available to set the location of the docker build context. Which is now set by `location:` key. However for backwards compatibility the previous usage still works, and will continue work (it is never going away). In fact, all usages of the `location:` + `dockerfile:` keys will be smartly interpreted. And you may set either one or both keys.
 
 ### `push`
 Maps to `docker push`.
@@ -69,7 +73,8 @@ The configuration defines a map of containers in either JSON or YAML. By default
 The map of containers consists of the name of the container mapped to the container configuration, which consists of:
 
 * `image` (string, required): Name of the image to build/pull
-* `dockerfile` (string, optional): Relative path to the Dockerfile
+* `location` (string, optional): Docker build context `PATH | URL`. Includes `Dockerfile`.
+* `dockerfile` (string, optional): Path to the Dockerfile. *[Note](#provision)*
 * `run` (object, optional): Parameters mapped to Docker's `run` & `create`.
 	* `add-host` (array) Add custom host-to-IP mappings.
 	* `cap-add` (array) Add Linux capabilities.
@@ -128,7 +133,7 @@ containers:
 			link: ["mysql:db", "memcached:cache"]
 			detach: true
 	app:
-		dockerfile: app
+		location: app
 		image: michaelsauter/app
 		run:
 			volume: ["app/www:/srv/www:rw"]
--- a/crane/container.go
+++ b/crane/container.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
 )
@@ -15,6 +16,7 @@ import (
 type Container interface {
 	Name() string
 	Dockerfile() string
+	Location() string
 	Image() string
 	Id() string
 	Dependencies() *Dependencies
@@ -42,6 +44,7 @@ type container struct {
 	id            string
 	RawName       string
 	RawDockerfile string          `json:"dockerfile" yaml:"dockerfile"`
+	RawLocation   string          `json:"location" yaml:"location"`
 	RawImage      string          `json:"image" yaml:"image"`
 	RunParams     RunParameters   `json:"run" yaml:"run"`
 	RmParams      RmParameters    `json:"rm" yaml:"rm"`
@@ -150,6 +153,10 @@ func (c *container) Name() string {
 
 func (c *container) Dockerfile() string {
 	return os.ExpandEnv(c.RawDockerfile)
+}
+
+func (c *container) Location() string {
+	return os.ExpandEnv(c.RawLocation)
 }
 
 func (c *container) Image() string {
@@ -403,7 +410,7 @@ func (c *container) Status() []string {
 }
 
 func (c *container) Provision(nocache bool) {
-	if len(c.Dockerfile()) > 0 {
+	if len(c.Dockerfile()) > 0 || len(c.Location()) > 0 {
 		c.buildImage(nocache)
 	} else {
 		c.pullImage()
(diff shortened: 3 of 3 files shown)


michaelsauter:
Just a quick note. Overall I'm not to happy with 2 separate names that do roughly the same thing. Moreover, I the term `location` is too general.

I do not fully understand why we need a new key to make this more flexible. What exactly prevents you from extending `dockerfile`?

dreamcat4:
OK @michaelsauter. I was hoping you would ask for some clarification about this. There is some subtelty to it. Much of the missing info can be gleaned from the `man docker-build` manpage.

But I shall try my best to clarify what is said there (as the manpage docs is a bit lacking). To explain what's up here, in my own words. Sorry if it's a bit long-whinded the explanations. But it should provide enough information. Please be sure to first read carefully the `docker-build` manpage too. As that is where most of this stuff is coming from...

* The `docker build` command takes 2 parameters. If only 1 parameter was needed then it would take only 1 parameter.

*However:*

*  *Only 1 of those 2 parameters is madatory for the basic case*.
* Where *the basic case* is when there is only 1 `Dockerfile`, which is placed in *the standard location*. Which is *the root folder of the build context.
* Where the *build context* is the mandatory argument, denoted by `PATH | URL` in the `man docker-build` manpage.
* And *the build context* is not necessarily only a local folder. It can be a remote URI resource, such as a git repository. Or it can be a tarball of files and folders.

So 1 key is fine to specify the build context. I call this key `location` in the source code of the pull request. BUT you can call it whatever you wish. However unfortunately the existing name `dockerfile` is not such an appropriate name for it anymore... beacuse since then the `docker build` command now has an optional parameter `--file=` to specify the dockerfile. So to avoid consfusion - that should be used for that purpose. And not to set the build context.

You say you don't like `location` because it is too vague. That is fine by me - I agree that aspect could be improved upon. Any reasonable alternatives, which may include `context` or `build_context`. You just have to choose a better name for it. Done / accepted.

Now we need to get to the part about having 2 different keys. We need 2 keys when we want to specify *variant images / multiple docker images based on the same build context*. A good example of such a user need is when you have one release build docker image for production. And then one variant used for debugging and developement. For example my Tvheadend image, has a debugging variant. Which is setup to include the nightly version of the software, and a binary with debugging symbols, the `gdb` application, and some scripts script to automatically take and save core dumps and stacktraces when the program crashes.

Yet the debugging image is nearly the same as the release image. So it needs to copy in all the same configuration files etc. And this is many files. So we need both images to be built from the same shared build context folder. Just to have `dock build` point to a different `tvh-debug.dkr dockerfile inside that same build context folder.

So that is what the `docker build --file=` optional argument is meant to be used for. Where the `PATH | URL` last argument is meant to remain the same.

Now this is where those hidden subtleties come into play:

* In cases where the `build context` is a single-layer local regular folder. Then we can automatically extract the last path component of the `dockerfile` key to determine the dockerfile. We are at the same time also extracting / discerning the location of the build context folder too.

e.g. `dockerfile: a_folder/app_variant.dkr`

`build_context` ---> `a_folder/`
`--file=` ---> `app_variant.dkr`

So then a single key works for that 1 situation. However there are now several situation where it does not work. Because the simple `basename` + `dirname` algorithm cannot auto-resolve those 2 combined components when:

* There is a nesting level of subdirectories. That is a real problem. Because I have seen in the wild on more than 1 occasion do this:

`build_context` ----> `my_folder/`
`dockerfiles` -----> `my_folder/dockerfiles/`
`--file=` ----> `myfolder/dockerfiles/app_variant.dkr`

So in that instance, there would be a bug in crane, because the build context would be extracted as `myfolder/dockerfiles`. And that would only contain just some dockerfiles and nothing else.

* Any situation when the `build_context` is not a local folder. This the `docker build` cmd that includes: tarballs, and other remote URI specified online resources. Most popularly including github (and non-github) git repos.

For example:

`build_context` ---> `https://github.com/account/repo.git`
`--file=` ---> `app_variant.dkr`
OR
`--file=` ---> `dockerfiles/app_variant.dkr`
etc.

In that circumstance. Then we cannot use simple file tests to auto-resolve and such combined context + dockerfile single key component. Because then to solve would require some problematic and very bad URI scheme(s) that are not syntactically correct / don't make sense.

e.g.

`dockerfile: https://github.com/account/repo.git/app_variant.dkr`
OR
`dockerfile: https://some.url.com/build_context.tar.gz/app_variant.dkr`

Which are both entirely invalid and can not be supported.

OR
You end up needing to to extra (even more) work, in order to peek at files through some github-specific URLs that can never work for regular (non-github) git repos.

So all of those cases are bad things - and can't be supported without a 2nd key.

Summary:
=========
Here is my suggestion / recommendations regarding this feature:

* As stated previously - we can easily rename `location:` to something else. Just tell me what you want instead of that and I can push another commit.
* If you don't wish any risk to confuse people away from established / entrenched `dockerfile:` key. Which actually has all along been setting the build context. We can not document at all the new 2nd key at all, for now, or at least the time being. OR if you consider this a breaking API change then designate the feature for Crane version 2. (however it does not break the existing API though).
* Or if you do want a documentation. But that I have proposed is simply is not good enough or up to scratch yet. Then please direct me further how to improve documentation aspect.

* It is just a suggestion - but perhaps don't be so keen to refuse features (like this one) when they put the `crane` tool a step behind and at a disadvantage to the official `docker compose` tool. Because now docker compose now officially has this feature. I believe they implemented it with seperate 2 keys too. Except they never called their original key for the build context `dockerfile:` to begin with - saving them some extra headaches.

Finally I also would like to state my commitment to code contributions: If you take my commit I will stand by it for any future maintenance or bug fixes. Just @me.  But also: I have thouroughly tested and checked through every line of this code several times over. It should not have have bugs in it, or require future revision / modification. However if the docker team improve their official `docker build` command. And that is not terribly likely, expect to support additional URI schemes. Then would be happy to come back if needed to re-service any such affected code. But it is not be needed and any future changes if docker add new URI schemes for build context...

dreamcat4:
Footnote:

One more choice for naming such a key would be `build:`. Since that is what fig / compose tool uses. So a `build:` key would require least explanation in docs and make some immediate sense to all people switching over, coming from fig / compose tool.

Personally I don't mind and have no preference over the key name.

BTW:

Technically, it is *possible* to get away with multple dockerfile (and "bug free") using only 1 key. However  you would need (as mentioned previously) to adopt some nonstandard custom syntax, denoting some special character in the middle to split up the 2 components. With crane to parse and split the string on.

*I do not recommended doing this.* However for completeness - such a method would look like this to users:

```yaml
# Seperator character: | (pipe)

dockerfile: build_context|dockerfiles/app_variant.dkr
OR
dockerfile: https://github.com/account/repo.git|dockerfiles/app_variant.dkr`
```

Of course this above ^^ theoretical '1-key' but-free method was not previously considered. So it has not been implemented in this PR.

Yet I would still generally recommend 2 keys approach here irrespective of the status of implemented code. As using 2 keys is the way most consistent with both official `docker build` cmd and official `docker compose` tool.

michaelsauter:
Sorry for not responding to this one for so long.

Basically, this is not a 100% finished in my eyes and I personally neither have a need for this nor think that Crane should allow all sorts of "build use cases". Crane's build functionality is quite limited, and I think that is okay. So please understand that I intend to focus on other things at the moment.

However, one takeaway from this pull request is that the `dockerfile` key that currently exists is not named very well. Maybe we should take the upcoming version 2 as a chance to rename it to `build`?

Then we can introduce a second `docker file` (or similarly named) key later on which would not introduce a breaking change compared to 2.0. Does that sound like a plan @dreamcat4?

dreamcat4:
@michaelsauter I would be simply ecstatic. Certainly a major release would be an appropriate time to switch the key name (if ever / else never). Many thanks. Appreciate it.

& no need to commit to anything else (beyond that) at this time.

michaelsauter:
Something else that came to my mind ... how about this structure:

```
containers:
  foo:
    build:
      context: some-directory
      file: Dockerfile.2
    run:
      ...
```

Then `build` would mirror the config for other commands like `run`. For 2.0, I would only implement `build/context` (keeping current functionality). Then at a later stage, we can add a `file` key.

dreamcat4:
Sure. I would be very pleased to see us move to that in crane 2.0.
It's better than my idea was :)

Perhaps something else to think about:

Do we need to print out some specifically-worded error msg when the previous `dockerfile:` key is encountered? I mean if this is a breaking change (i don't know if it is).

An informative message can be less burdensome to people if it clearly explains how to fix their `crane.yml` file(s) with the new key name. When upgrading from crane v1.x. It only needs to be done 1 time for everybody.

Else we may allow lazy adoption of new keys. By permitting the old `dockerfile:` key in there for some months as a chageover period. The old key would still remain valid / honoured. But whenever crane runs it prints out a warning message. Telling people they are using a deprecated key now.

If both keys are detected simultansously, then I suppose we aught to just completely ignore the old one.

Mostly the action taken depends upon how much finnikey error handling / quirk behaviour you permit in your code. They are each valid choices / options.

Including generic error message ('invalid key') which is not customized whatsoever. (the least effort option).

PR #173, opened 2015-04-03
Templating feature

adrianhurt:


adrianhurt:
Hi, this PR allows you to define your own template (and not only the DOT one) to generate an output file (or a set of files) from your information on crane.yml.

If you agree with this, I will add the tests and docs. But for the moment is a draft.

As I said, for the moment I haven't documented it, but in summary, it's a generalized feature for the current graph option, but with your own template. There are 2 arguments:

* Input template file (`-t`, `--tmpl`). Required.
* Output file (`-o`, `--output`). Optional. If it's not present, the output will be printed in the console (`os.Stdout`). If it's present there are 2 options:
  * Regular file name (ex: _myGraph.dot_): the output will be saved in that file.
  * File format containing `%s` (ex: _%s.service_): it will generate a file for each container applying the same template for all of them. This format determines the name of each file.

I've included some template examples for _DOT graph_, _Systemd_ and _Upstart_. In my case I've used the _Systemd_ example in _CoreOS_ and works perfectly taking into account all the dependencies and generating every _service_ file.

To use the examples:
```
crane template -t path/to/graph.tmpl -o output/myGraph.dot
crane template -t path/to/systemd.tmpl -o output/%s.service
crane template -t path/to/upstart.tmpl -o output/%s.conf
```

crane/template.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+type TemplateData struct {
+	Graph              DependencyGraph
+	TargetedContainers Containers
+}
+type TemplateDataForContainer struct {
+	Graph              DependencyGraph
+	TargetedContainers Containers
+	CurrentContainer   Container
```
bjaglin: I guess we should reduce the amount of stuff exposed in `Container` - side-effect functions like `Run()` or `Kill()` should not be available.

crane/template.go
```diff
@@ -0,0 +1,105 @@
...
+	"strings"
+	"text/template"
+)
+
+type TemplateData struct {
+	Graph              DependencyGraph
+	TargetedContainers Containers
+}
+type TemplateDataForContainer struct {
+	Graph              DependencyGraph
```
bjaglin: might be worth considering exposing `Dependencies` as an interface rather than a struct for the sake of abstraction

crane/cmd.go
```diff
@@ -201,7 +211,7 @@ in the config, but not defined). Targeted containers are highlighted with color
 borders. Solid edges represent links, dashed edges volumesFrom, and dotted edges
 net=container relations.`,
 		Run: configCommand(func(config Config) {
-			config.DependencyGraph().DOT(os.Stdout, config.TargetedContainers())
+			ConfigToTemplateData(config).DOT(options.tmplOutput)
```
bjaglin: I guess you can take out the old function in DependencyGraph() then

crane/cmd.go
```diff
@@ -252,6 +262,11 @@ See the corresponding docker commands for more information.`,
 	cmdLogs.Flags().StringVarP(&options.tail, "tail", "", "all", "Output the specified number of lines at the end of logs")
 	cmdLogs.Flags().BoolVarP(&options.colorize, "colorize", "z", false, "Output the lines with one color per container")
 
+	cmdTmpl.Flags().StringVarP(&options.tmplInput, "tmpl", "t", "", "The input template file")
```
bjaglin: why not spell `template` out completely?

crane/cmd.go
```diff
@@ -252,6 +262,11 @@ See the corresponding docker commands for more information.`,
...
+	cmdTmpl.Flags().StringVarP(&options.tmplOutput, "output", "o", "", "The output file or directory (os.Stdout as default)")
```
bjaglin: how would directory work?

bjaglin:
:+1: I personally love it, great idea! I guess we just need to agree on what's available in the template - the current code isn't particularly thought for being used externally.

crane/template.go
bjaglin: Also, I guess the original `Graph`/`DependencyGraph` naming is a bit off, since it's a structure representing a graph (which happens to be a tree) but expressed "flat".

crane/template.go
adrianhurt: Yes! maybe called `ContainerData`?
```
type ContainerData interface {
	Name() string
	Dockerfile() string
	Image() string
	Id() string
	Dependencies() *Dependencies
	Exists() bool
	Running() bool
	Paused() bool
	ImageExists() bool
	Status() []string
}
```

crane/template.go
adrianhurt: For `Dependencies`, do you mean to create an interface mapping the current struct? Called `Dependencies` as well?

And for `Graph`, I think I understand your point… maybe `DependencyMap`?

crane/cmd.go
adrianhurt: Yes, of course. I intended avoid to touch the rest of the code for the draft.

crane/cmd.go
adrianhurt: Ok, it was simply for avoiding `crane template —template "file.tmpl"`. It seems too redundant.

crane/cmd.go
adrianhurt: Oh, ignore that, sorry. It's because previously I thought in indicating a directory to output every file with the name of each container, but then I decided to use a string format with `%s` to be able to adding a file extension. I will check every text, I only was waiting to the PR is approved.

adrianhurt:
Thanks! And yes, the idea it was already there for DOT graphs. It's only a little refactoring to do it more general.

adrianhurt:
BTW, I added some utility functions for the templates. I chose them because I think they could be the most useful ones, but maybe I forgot someone else. There are:
* strings.Contains
* strings.ContainsAny
* strings.HasPrefix
* strings.Join
* strings.Split
* strings.ToLower
* strings.ToTitle
* strings.ToUpper
* regexp.MatchString

README.md
```diff
@@ -192,6 +195,74 @@ This could be used like so: `crane provision service1`, `crane run -v databases`
...
+
+* __Generate a file per container:__ Specify a file format containing `%s` to use the template to generate a different output file for each container using the same template (substituting the `%s` by the corresponding container's name).
+```
+crane template -t path/to/myTemplate.tmpl -o output/%s.txt
+```
+
+The data object passed to the template is a `TemplateInfo` struct, or a `TemplateInfoForContainer` one if you want to generate a file per container. You can see here the structure of each struct or interface needed.
+
+```go
+type TemplateInfo struct {
```
bjaglin: This might get outdated - a link to https://godoc.org/github.com/michaelsauter/crane/crane might be better to avoid keeping that in sync?

crane/test_templates/test.tmpl
```diff
@@ -0,0 +1,17 @@
...
+    image: {{ .Image }}
+    id: {{ .Id }}
+    dependencies: {{ Join .Dependencies.All ", " }}
+    exists: {{ .Exists }}
+    running: {{ .Running }}
+    paused: {{ .Paused }}
+    imageExists: {{ .ImageExists }}
+    status: [{{ Join .Status ", " }}]{{ end }}{{ end }}
+
+DEPENDECY-MAP:{{ range $name, $dependencies := .DependencyMap }}
```
bjaglin: typo - `DEPENDENCY-MAP`

crane/template.go
```diff
@@ -0,0 +1,131 @@
...
+				return
+			}
+		}
+	} else {
+		generateFile(tmpl, outputPath, templateInfo)
+	}
+}
+
+// generates a file with the output. If outputPath is empty, it will be dumped to os.Stdout
+func generateFile(tmpl *template.Template, outputPath string, templateInfo interface{}) error {
```
bjaglin: no client seems to care about that error returned

crane/cmd.go
```diff
@@ -252,6 +268,11 @@ See the corresponding docker commands for more information.`,
 	cmdLogs.Flags().StringVarP(&options.tail, "tail", "", "all", "Output the specified number of lines at the end of logs")
 	cmdLogs.Flags().BoolVarP(&options.colorize, "colorize", "z", false, "Output the lines with one color per container")
 
+	cmdTmpl.Flags().StringVarP(&options.tmplInput, "template", "t", "", "The input template file name")
```
bjaglin: continuing here the discussion about `template` as the long parameter name... I see your point about the fact that it's weird to have both the command and a flag with a similar key... Maybe that calls for a different command name - `generate`, `transform` ?

crane/template.go
```diff
@@ -0,0 +1,131 @@
...
+	"fmt"
+	"github.com/michaelsauter/crane/print"
+	"os"
+	"path"
+	"regexp"
+	"strings"
+	"text/template"
+)
+
+type TemplateInfo struct {
```
bjaglin: Maybe I am too defensive and internal interfaces shouldn't be public as they are now, but I feel all exposed structures should be subset interfaces of the internal ones - so it's clear that we break the external contract if we remove/update an internal function/struct/whatever.

crane/template.go
```diff
@@ -0,0 +1,131 @@
...
+
+// returns a TemplateInfoForContainer from a ContainerInfo and a container
+func (td TemplateInfo) forContainer(currentContainer ContainerInfo) TemplateInfoForContainer {
+	return TemplateInfoForContainer{currentContainer, td.TargetedContainers, td.DependencyMap, td.Groups}
+}
+
+// generates a textual output for the selected template and templateInfo
+func (templateInfo TemplateInfo) template(templatePath string, outputPath string) {
+	if templatePath == "" {
+		print.Errorf("You must select the input template with the parameter --template (-t)\n")
```
bjaglin: shouldn't that be enforced earlier? It feels weird to see hardcoded CLI stuff here

crane/cmd.go
```diff
@@ -192,6 +193,21 @@ formatted in bold provided the 'colorize' flag is on.`,
 		}, true),
 	}
 
+	var cmdTmpl = &cobra.Command{
+		Use:   "template",
+		Short: "Generates a textual output using Go templates",
+		Long: `Generates a textual output using Go templates. Use the 'template' flag to
+indicate the input template. Depending on the optional 'output' flag's value the
+behaviour is different. If you leave it, the output will be dumped through Stdout.
+If you specify a filename, the output will be rendered on that file. Additionally,
```
bjaglin: `Additionally` makes me think that `the output will be rendered on that file` remains true in the `%s` case, which is not the case, it really is an either-or.

templates/readme.md
```diff
@@ -0,0 +1,61 @@
+# DOT Graph template
+
+```
+crane template -t path/to/graph.tmpl -o output/myGraph.dot
+```
+
+Generate a [DOT](http://en.wikipedia.org/wiki/DOT_(graph_description_language)) file representing the dependency graph. Bold nodes represent the containers declared in the config (as opposed to non-bold ones that are referenced in the config, but not defined). Targeted containers are highlighted with color borders. Solid edges represent links, dashed edges volumesFrom, and dotted edges net=container relations.
+
+# Systemd template
```
bjaglin: couldn't these templates be tested from `template_test.go` to make sure they remain valid no matter what we do (horrible conscious breaking changed or other) ?

crane/template.go
```diff
@@ -0,0 +1,131 @@
...
+	Groups             map[string][]string
+}
+
+// Sub-interface of Container
+type ContainerInfo interface {
+	Name() string
+	Dockerfile() string
+	Image() string
+	Id() string
+	Dependencies() *Dependencies
```
bjaglin: I know this makes life easier when generating a per-container systemd/upstart file, but I feel like this shouldn't be exposed as-is here. I see `ContainerInfo` as something providing either raw configuration, or state about that container. `Dependencies` is an opinionated transformation of  links/volumesFrom/net, and is already available via a lookup in the `DependencyMap`. If we need access to raw dependencies, maybe we could expose `RunParams()` ? I know it seems clunky but what if we need access to the exposed ports from the template?

bjaglin:
Just a heads up regarding my comments - @michaelsauter still hasn't given his opinion about whether/ how should go in, so take them with a grain of salt.

README.md
adrianhurt: Yes, I agree. Particularly https://godoc.org/github.com/michaelsauter/crane/crane#TemplateInfo and https://godoc.org/github.com/michaelsauter/crane/crane#TemplateInfoForContainer

crane/test_templates/test.tmpl
adrianhurt: Thanks

crane/template.go
adrianhurt: I had 
```
err = generateFile(tmpl, outputPath, templateInfo)
if err != nil {
  return
}
```
But at the end of the function it has no effect in that point. And the error is already printed inside the function

crane/cmd.go
adrianhurt: I see your point. Now I'm not sure which name should have the feature… With `template` I understand better what it is intend, and the other ones are more ambiguous. But the action itself is not template but the file used to generate an output. So I would prefer `generate` for this point. The problem I see for using `generate` is that it's a very generic word.

crane/template.go
adrianhurt: Maybe

crane/template.go
adrianhurt: Yes, you're right. In deed, it was a question I would like to ask you and I forgot. How can I force a flag to be present (with Cobra)?

crane/cmd.go
adrianhurt: Right, then `Additionally` -> `Finally`?

templates/readme.md
adrianhurt: Yes, but I think the idea should be that these are only examples. Crane shouldn't support them… I can add more templates in the `test_templates` folder for them.

crane/template.go
adrianhurt: I see your point with this as well. We could remove `Dependencies` from `ContainerInfo` and check them directly from `DependencyMap`.

For the other hand, it's true that making `RunParams()` accessible could be interesting. But then `RmParams()` and `StartParams()` as well?

In fact, I was thinking to give all the information of `config` struct, but interface `Config` limit us now, so I would need to add some methods. I mean to have something like this:
```
type TemplateInfo struct {
  ContainerMap   map[string]ContainerInfo
  DependencyMap  DependencyGraph
  Target         []string
  Groups         map[string][]string
}

func (ti TemplateInfo) TargetedContainers() []ContainerInfo
func (ti TemplateInfo) Dependencies(c ContainerInfo) *Dependencies
func (ti TemplateInfo) ContainersForGroup(g string) []ContainerInfo
func (ti TemplateInfo) IsTarget(c ContainerInfo) bool
func (ti TemplateInfo) BelongsToGroup(g string) bool
// you can think more...
```

adrianhurt:
Ok, let's wait to @michaelsauter

templates/readme.md
bjaglin: I understood this was given as an example/suggestion, but I just hate to see documentation examples which don't work, so running them as part of the automated testing would prevent that.

crane/template.go
bjaglin: I don't think you can, you'd have to add a checker function called from all command handlers utilizing this option. Not perfect, but as long as it's contained together with the other CLI stuff, it's not too ugly.

crane/template.go
bjaglin: then don't return an error at all? but shouldn't an error be fatal at this point?

adrianhurt: with a `panic(StatusError{fmt.Errorf(errorDescription, errorStatus})`? Which `errorStatus` can I use? There is any convention?

crane/template.go
adrianhurt: Something like this?
```
func requiredStringFlag(name string, value string) {
  if value == "" {
    panic(StatusError{fmt.Errorf("The flag '%s' is required", name), XX})  //  XX = 64??
  }
}
...
Run: configCommand(func(config Config) {
  requiredStringFlag("template", options.tmplInput)
  ConfigToTemplateInfo(config).template(options.tmplInput, options.tmplOutput)
}, true),
...
```

michaelsauter:
@adrianhurt @bjaglin Thanks for your great work on this. I love the idea in general, so please go ahead with it. I had a quick look through the code and found it a bit difficult to track, but it's quite a big change so that's to be expected. Since I don't have much time at the moment, I will leave this with you guys if you're okay with it. Once you think it's done, I can do a final walk-through and merge it.

adrianhurt:
Sorry to be disappeared for so much time. I'll try to finish this in a few days when I get some free time.

jleskovar:
Hi @adrianhurt 
Just wondering where you guys were at with the PR?
Cheers

michaelsauter:
Closing this for now due to inactivity. Feel free to reopen after 2.0 is released.

michaelsauter:
I'm not sure if you're still interested in this, but I loved the idea of this. Finally got around to implement it, basing it somewhat loosely on this PR.

The main commits are:
* 4e7c786136bac7dbd4f4aeb7fa68c2c912ec5470 extracts an interface with basic container information for use in templates
* 73b11071ae0e12e9484375c6309c88d20d14228a adds the actual command

PR #176, opened 2015-04-08
Add default-env parameter

williamhogman:
Hello there!

First of, thanks for an awesome project, as a developer using docker daily crane really makes my life easier.

This commit adds a parameter by the name ```env-default```. The purpose of ```env-default``` is to provide a means by which developers can set reasonable defaults in crane.yaml (e.g development settings) while still allowing these to be overridden manually using environment variables. While it is certainly possible to achieve this by other means they would require either preprocessing crane.yaml or wrapping crane with another tool setting environment variables, which seems needlessly complex.

Finally, I wasn't sure about unit-tests so I haven't included any. I'm happy to write them if you could point me towards an existing test-case in the code-base that I could adapt to my code.

All the best,
/will

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ The map of containers consists of the name of the container mapped to the contai
 	* `dns` (array)
 	* `entrypoint` (string)
 	* `env` (array)
+	* `env-default` (array) Default values for environment variables
 	* `env-file` (array)
 	* `expose` (array) Ports to expose to linked containers.
 	* `hostname` (string)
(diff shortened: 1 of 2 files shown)


bjaglin:
The `run` block is currently a 1-1 mapping with docker flags and will remain so, so that would need to live in another block. But I must say I don't get your use-case, could you post an example - basic `crane.yml` and a worflow? I

michaelsauter:
@williamhogman Thanks for sharing this. As @bjaglin already said, `run` will stay a 1:1 mapping.

Regarding your use case, I actually think environment variables are a great way to manage environment specific stuff. Maybe you could solve your problem easily by moving all your variables into a file (and giving that to Crane with `env-file`), and then just swapping the filename with one environment variable that needs to be set?

williamhogman:
@michaelsauter  That's a great idea but I don't think it quite fits our use-case. 

Basically, and I will get back to you guys with a full example and what not in a bit, what we are trying to do is this. We have a development environment where a bunch of docker containers link to each other which means a bunch of defaults like ```DATABASE_URL=postgresql://somelinked-container``` when we roll this out in production we want another value for DATABASE_URL. Right now we basically use bash for templating our crane.yml file and have a bunch of ```${DATABASE_URL-postgresql://somecontainer}``` but this feels awfully hacky.

If you have guys have any ideas on how to implement something like this in a way that fits the ethos and style, I'd be happy to pitch in. :)

michaelsauter:
@williamhogman Hmm. I don't quite get why 2 different environment files don't work for you then? You just have 2 files with diferent values. One for dev, one for prod.

williamhogman:
@michaelsauter We figured out a way of solving that didn't involve having to use bash for templating crane.yml. Thanks a ton for the help :+1:

Issue #177, opened 2015-04-17
Support for multiple link aliases


mishak87:
It would be nice to support [multiple links of single container](https://github.com/docker/docker-py/pull/153).

It is useful ie. for application that has several databases but for dev it is simpler to use just one.
db1.host db2.host, db1-ro.host etc. (example taken from legacy app)

Defining more aliases results in `ERROR: runtime error: slice bounds out of range`.
```yaml
containers:
    a:
        run:
            link:
                - db:db1.host
                - db:db2.host
                - db:db1-ro.host
```

PR #178, opened 2015-04-17
Support for multiple links to same container #177

mishak87:
Well this was rather easy :)

### DIFF
--- a/crane/container.go
+++ b/crane/container.go
@@ -122,8 +122,10 @@ func (c *container) Dependencies() *Dependencies {
 	dependencies := &Dependencies{}
 	for _, link := range c.RunParams.Link() {
 		linkParts = strings.Split(link, ":")
-		dependencies.All = append(dependencies.All, linkParts[0])
-		dependencies.Link = append(dependencies.Link, linkParts[0])
+		if !dependencies.includes(linkParts[0]) {
+			dependencies.All = append(dependencies.All, linkParts[0])
+			dependencies.Link = append(dependencies.Link, linkParts[0])
+		}
 	}
 	for _, volumeFrom := range c.RunParams.VolumesFrom() {
 		if !dependencies.includes(volumeFrom) {
--- a/crane/container_test.go
+++ b/crane/container_test.go
@@ -9,12 +9,19 @@ import (
 
 func TestDependencies(t *testing.T) {
 	c := &container{RunParams: RunParameters{RawNet: "container:n", RawLink: []string{"a:b", "b:d"}, RawVolumesFrom: []string{"c"}}}
-	if deps := c.Dependencies(); deps.All[0] != "a" || deps.All[1] != "b" || deps.All[2] != "c" || deps.All[3] != "n" || deps.Link[0] != "a" || deps.Link[1] != "b" || deps.All[2] != "c" || deps.Net != "n" {
+	if deps := c.Dependencies(); deps.All[0] != "a" || deps.All[1] != "b" || deps.All[2] != "c" || deps.All[3] != "n" || deps.Link[0] != "a" || deps.Link[1] != "b" || deps.Net != "n" {
 		t.Errorf("Dependencies should have been a, b, c, n. Got %v", deps)
 	}
 	c = &container{RunParams: RunParameters{RawLink: []string{}, RawVolumesFrom: []string{}}}
 	if deps := c.Dependencies(); len(deps.All) != 0 && len(deps.Link) != 0 && deps.Net == "" {
 		t.Error("Dependencies should have been empty")
+	}
+}
+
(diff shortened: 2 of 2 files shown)


bjaglin:
Thanks for this! Could you add the scenario described in #177 as a test to prevent regressions and document that behavior?

mishak87:
Will do this evening.

mishak87:
OT: I have few suggestions to keep code simple. Subdirectories seem unnecessary was there any reason for spliting code in them?
Also would be best to setup travis CI for tests :)

michaelsauter:
@mishak87 Please open a new ticket regarding code org. Not sure what you mean though :)

mishak87:
@michaelsauter Sorry that was beside the point, I will prepare example pull request for that stuff. Will be faster to explain that way. :)

Any suggestions to this pull or does it LGTY?

michaelsauter:
LGTM, please squash the commits into one and I will merge it in. Thanks!

crane/container.go
```diff
@@ -122,7 +122,9 @@ func (c *container) Dependencies() *Dependencies {
 	dependencies := &Dependencies{}
 	for _, link := range c.RunParams.Link() {
 		linkParts = strings.Split(link, ":")
-		dependencies.All = append(dependencies.All, linkParts[0])
+		if !dependencies.includesAsKind(linkParts[0], "link") {
```
bjaglin: While you are it, I think you should rather follow the same pattern as below for VolumesFrom, to avoid having duplicates anywhere - namely:
```
		name = linkParts[0]
		if !dependencies.includes(name) {
			dependencies.All = append(dependencies.All, name)
			dependencies.Link = append(dependencies.Link, name)
		}
```

crane/container_test.go
```diff
@@ -18,6 +18,13 @@ func TestDependencies(t *testing.T) {
 	}
 }
 
+func TestMultipleLinks(t *testing.T) {
```
bjaglin: how about `TestMultipleLinkAliases` ?

crane/container.go
mishak87: I don't think supporting volumes-from will make sense. Is there any possible use-case for that in docker?
But it would be better to have more user friendly error message then "index out of bounds".

bjaglin: VolumesFrom is already supported, just below. Have a look at it, my point was just that if we de-duplicate `.All`, we might as well deduplicate `.Link` (like we do for `.All`/`.VolumesFrom`) by guarding them within the same `if` block.

mishak87: I don't think we understand each other. My assumptions are:
1st) `--link` can be used to create multiple aliases. Mathematically its effect is similar to addition.

2nd) There is no point to support multiple `--volumes-from` of same container. It would result in mapping volumes over mapped volumes of that same container. Mathematically it resembles multiplication.

3rd) Since `--link` is an addition it must be handled specially. Using same pattern as for `--volumes-from` would result in addition of only the first `--link` occurence for each container.
That would prevent the error, but is undesired since it would also disable the feature I am trying to use.

Please correct me if I am wrong!

bjaglin: Agree with all your points, but I think it's not relevant to our discussion - I think you misinterpreted the semantics of `Dependencies` (probably due to a lack of documentation!).

The actual usage of multiple aliases does not rely on `Dependencies`, but on `Links()` https://github.com/michaelsauter/crane/blob/0e58aa4202647770d2f4484876331712369a8f8a/crane/container.go#L533-L536.

If there was a built-in `Set` in golang, members of `Dependencies` would be of that type - there is no benefit in having duplicated entries (as you have noticed by hitting the bug in the first place) - check throughout the code, it's only used to append and check for existence/"contains". So as you were about to prevent duplicate in the `.All` member, I thought you could also do it in `.Link`, as it's done for the other source of dependencies such as `--volumes-from` or `--net container:<foo>`

mishak87: **TL;DR** Fixed. Thank you for your patience! :smiley:

I had to re-test it manually and realised, that I have made mistake while testing the code for the first time (10 days ago). I have considered and tried your variant first, but probably I have built binary for wrong code and it did not work. Since then I had assumed it was not working due to internal logic. Sadly I did not revise the premise when I got more familiar with the code.

It might help to have automated tests in travis to prevent such issues in future along with CI for building binaries.

bjaglin: +1 for the CI, it's quite easy to set up, I'll open a ticket as a reminder

bjaglin:
LGTM (although @michaelsauter will prefer a squash of your 3 commits)

mishak87:
Squashing away... :smiley:

michaelsauter:
:+1:

PR #179, opened 2015-04-20
Adding support for env as a mapping

adrianhurt:
Sorry, I needed to redo the PR. This is a copy of the #168. But now it's been rebased to be ready to merge.

It allows to declare the env as a mapping in addition to the array syntax. It lets declare alias in the yaml file and use it as variables and also merge blocks easily.

See more and examples in issue #163 (#163)


michaelsauter:
Awesome, thanks!

PR #182, opened 2015-04-30
use testify for assertions in unit tests

bjaglin:
Current unit tests are very hard to read, maintain and debug. This aims at improving them by using a simple assertion library. This PR should be a no-op - assertions/contracts remain the same.

### DIFF
--- a/crane/config_test.go
+++ b/crane/config_test.go
@@ -1,7 +1,7 @@
 package crane
 
 import (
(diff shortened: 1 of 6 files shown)


michaelsauter:
Wow. That is so much better. Thanks heaps!

michaelsauter:
Oops, I didn't merge it.

Issue #166, opened 2015-03-25
boot2docker support


mishak87:
- [ ] 386 binaries - it is run in 32b userspace
- [ ] use sh instead of bash for download script - no bash

mishak87:
I can provide pull request if necessary.

bjaglin:
Could you provide more details with the use case? I am not very familiar with boot2docker, but isn't Docker only available on x64 architecture - meaning that boot2docker is not 32bits?

mishak87:
https://github.com/boot2docker/boot2docker-cli/issues/264
> boot2docker currently uses Tiny Core Linux, which provides a 64-bit kernel and a 32-bit user-space.

bjaglin:
Oh, I wasn't aware of that userspace twist - thanks.

michaelsauter:
@mishka87 Pull request is welcome. I don't use boot2docker, but happy to merge the changes as they seem small enough.

mishak87:
Thanks will try to do it over the weekend.

mishak87:
Finally got the time to do it #184 and #185.

michaelsauter:
Closing this since #185 is merged already. Hesitant about #184 though, but that scripts needs to be rewritten anyway ...

PR #185, opened 2015-05-01
Linux 386 build #166

mishak87:
I finally got around to do this :smile:.


michaelsauter:
@mishak87 Could you remove the download script change from this PR?

Issue #181, opened 2015-04-28
Set up CI hooks


bjaglin:
As the number of contributors increase, we should have proper build status on PRs and on the README.md. I have done it quite easily in https://github.com/bjaglin/multiplexio for example.

bjaglin:
cc @mishak87

michaelsauter:
Sounds good. I started doing it but the build failed. Will have to investigate ...

bjaglin:
I think it's related to some unit tests not so "unit", and requiring the `docker` client to be in the path. I am actually working towards fixing that in an unrelated PR where I need to stub it.

bjaglin:
Nevermind, the `docker` stubbing is not the issue (tests handle the absence of `docker` somehow). It's actually related to a breaking change in https://github.com/fatih/color/commit/85fc70c416e81ae0e3ef7913d586d9ea5c15a295 - I'll look at updating the code so that it builds with the HEAD of fatih/color.

michaelsauter:
Ah right, because it does not know about which versions we use. Yay for Go dependency management...

bjaglin:
It's working now - for open PRs, closing/reopening does the trick

PR #187, opened 2015-05-04
reflect breaking changes in fatih/color:85fc70c (color.Set() private)

bjaglin:
Simply allows to build Crane against latest `fatih/color` - no functional change.
Fixes https://github.com/michaelsauter/crane/issues/181 :sunny:


fatih:
Sorry again, this is imho something I did. I've totally forget about the publicity of `color.Set()`. I've reverted it again with https://github.com/fatih/color/pull/14. Please let me know if the there is something to do.

bjaglin:
OK, maybe let's hang on with this one - the breaking change is apparently going to be reverted (see issue above).

michaelsauter:
@fatih thanks for reverting!

PR #197, opened 2015-06-09
Search parent directories for config too

michaelsauter:
If none of the config files is found in the current dir, it continues to search the parent directories recursively.

I can't see why this would break, but maybe someone knows of a use case where this is unintended?

My main use case is that you might be in a sub-dir of your project and want to execute a Crane command (without specifying the file or changing to the root dir).

### DIFF
--- a/crane/config_test.go
+++ b/crane/config_test.go
@@ -2,7 +2,11 @@ package crane
 
 import (
 	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"sort"
+	"syscall"
 	"testing"
 )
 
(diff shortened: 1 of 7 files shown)


dreamcat4:
+1

bjaglin:
I feel like going up the filesystem hierarchy is a bit intrusive... To cover your main use-case, wouldn't an environment variable (`CRANE_CONFIG`, or `CRANEPATH` to mimic go convention?) be a better way to reference a configuration/directory via an absolute path, allowing to run `crane` no matter what your current `pwd` is?

dreamcat4:
Another env var could switch on/off the automatic `../` directory searching behaviour. But I think it's useful to provide *both of those things*.

There are similar features that git supports. Maybe crane can be a bit cleaner / simpler than git though. At any rate, I would find both options useful:

1) Is used when browsed in a subdirectory of the root crane folder (as is also common in git repo).

2) I imagine is useful when defining global shell alias like this:

```sh
alias crane-set1="CRANE_PATH=$HOME/set1 crane"
alias crane-set2="CRANE_PATH=$HOME/set2 crane"
```

Where presumably 2) you might be browsing anywhere else *outside them, in other folders* at the time when you need to invoke some crane cmds back in your own crane folder(s).

It seems like different situations from each other.

michaelsauter:
@bjaglin What concrete issues do you see with traversing upwards?

I don't like `GOPATH` at all, so as a consequence I don't really like an env var for this either:

a) It requires additional setup vs. working out of the box
b) There can only be one env var (without workarounds as described by @dreamcat4), which makes it hard to use with multiple crane backed environments (which I do have)

To me it makes sense to allow Crane to work "inside the project", similar to git.

I have 2 use cases:

a) being in a subfolder (e.g. in a rails project, being in `app/models` or so instead of the root where the crane config is)
b) having the crane file outside individual projects in a "dev env folder" (e.g. there are multiple subsystems. most devs work on just one subsystem, but some work on multiple ones. also, the subsystems my be connected via a message bus or similar. then it is convenient to have a folder with a crane file which defines the containers for the message bus and the subsystems. devs can clone the subsystems they need. however, since most work on only one system, it is convenient to be able to run crane commands from the root of their subsystem)

michaelsauter:
I've updated the PR with 2 changes/fixes:

a) When an absolute config path is given, no traversal happens (obviously)
b) When a relative path is specified in the run config `volume`, the path to the config file is used to translate it into an absolute path to be passed to `docker run`.

crane/config_test.go
```diff
@@ -24,17 +28,36 @@ func (stubbedContainer *StubbedContainer) Exists() bool {
...
-	files := configFiles(options)
+	files := configFilenames(options)
 	assert.Equal(t, []string{filename}, files)
 	// Without given filename
-	files = configFiles(Options{})
+	files = configFilenames(Options{})
 	assert.Equal(t, []string{"crane.json", "crane.yaml", "crane.yml"}, files)
 }
 
+func TestFindConfig(t *testing.T) {
```
bjaglin: maybe you can add a test verifying that an absolute path works?

bjaglin:
Of course... I never realized that's what git is doing! So forget my first comment, I have nothing against the feature - I understand better the use case now and the behavior with volumes feels like a no-brainer :+1:

crane/config_test.go
michaelsauter: Good point, added.

Issue #195, opened 2015-05-28
Thinking about Crane 2


michaelsauter:
Recently I have been thinking about what a version 2 of Crane could be. Parts of this has already been covered in #186, but I would like to collect more thoughts here - and hopefully get some feedback!

What I don't like about Crane currently:
* Mixture of build and runtime functionality. In #186 I was suggesting to extract the build related part, now I'm thinking I might actually just remove it altogether. I don't see it as a goal to have feature parity with Docker Compose, and what Crane currently offers in terms of building images is so small it can easily be handled by a script. The difficulty with extracting is that it would need to share quite a bit of the code, and that it would take time to maintain
* No way to specify CMD via the CLI. I would like to be able to specify a different process for the container(s). That is currently impossible since there can be multiple targets.
* Too complicated cascading (we probably don't need the different types)
* Odd behaviour of `lift` - I can't think of a use case where you wouldn't want to pass `-r` to it

Possible changes:
* Remove `provision` and `push` commands
* Simplify cascading to dependencies/affected. Possibly with a more fine-grained linked/linking since that seems to be the most common use-case
* Map `lift` to pull//remove/run
* Have only one target. That allows users to specify a process to run as the second argument
* Maybe that one target can be a bit more complex. E.g. `a+b` to target both `a` and `b`. One could also see affected containers as some sort of "dynamic group", which might be nice to specify as a target as opposed to via flags, e.g. `a+linked` or if we want to use a different char, `a&linked`. To exclude targets, something like `a^b` could work.

Any feedback welcome.

dreamcat4:
Hello @michaelsauter I just wanted to check that you are not planning on removing the `crane create` cmd? It should not be, yes?

michaelsauter:
@dreamcat4 Of course not, `create` is staying for sure. BTW, I still haven't decided on whether some build functionality should go or stay. At this stage, these are just thoughts, no concrete plans.

ncadou:
Right now I'm not using the build functionality because, as you said, it's too limited. The ability to specify a command as an alternative to a Dockerfile would go a long way in my case.

missedone:
maybe in V2, can consider use golang docker client, e.g. https://github.com/fsouza/go-dockerclient, so that we can get rid of the dependency of docker bin.
i can take a look if you're interested

dreamcat4:
@missedone Actually I like that crane calls the docker bin... because it means the `-v` switch shows exactly which docker commands are being run. That is very practical / useful.

[...]

PR #198, opened 2015-06-09
Hack to showcase passing commands to crane CLI

michaelsauter:
This is a hack to partially implement what I was thinking for #195.

Feedback and tests welcome.

### DIFF
--- a/crane/cmd.go
+++ b/crane/cmd.go
@@ -23,6 +23,7 @@ type Options struct {
 	ignoreMissing       string
 	config              string
 	target              []string
+	cmd                 []string
 }
 
 var options Options
@@ -42,8 +43,11 @@ func configCommand(wrapped func(config Config), forceOrder bool) func(cmd *cobra
 			}
 		}
 
-		// Set target from args
-		options.target = args
+		// Set target and potentially command from args
+		if len(args) > 0 {
+			options.target = strings.Split(args[0], "+")
+			options.cmd = args[1:]
+		}
 
 		config := NewConfig(options, forceOrder)
 		if containers := config.TargetedContainers(); len(containers) == 0 {
@@ -65,7 +69,7 @@ func handleCmd() {
 		Long: `
 lift will provision missing images and run all targeted containers.`,
 		Run: configCommand(func(config Config) {
-			config.TargetedContainers().lift(options.recreate, options.nocache, options.ignoreMissing)
+			config.TargetedContainers().lift(options.recreate, options.nocache, options.ignoreMissing, options.cmd)
 		}, false),
 	}
 
@@ -104,7 +108,7 @@ pull will pull the image(s) from the given registry.`,
 		Short: "Run the containers",
 		Long:  `run will call docker run for all targeted containers.`,
 		Run: configCommand(func(config Config) {
-			config.TargetedContainers().run(options.recreate, options.ignoreMissing)
+			config.TargetedContainers().run(options.recreate, options.ignoreMissing, options.cmd)
 		}, false),
 	}
 
--- a/crane/container.go
+++ b/crane/container.go
@@ -27,9 +27,9 @@ type Container interface {
 	ProvisionOrSkip(update bool, nocache bool)
 	PullImage()
 	Create(ignoreMissing string)
-	Run(ignoreMissing string)
+	Run(ignoreMissing string, cmd []string)
 	Start()
-	RunOrStart(ignoreMissing string)
+	RunOrStart(ignoreMissing string, cmd []string)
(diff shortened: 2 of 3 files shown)


crane/cmd.go
```diff
@@ -42,8 +43,11 @@ func configCommand(wrapped func(config Config), forceOrder bool) func(cmd *cobra
 			}
 		}
 
-		// Set target from args
-		options.target = args
+		// Set target and potentially command from args
+		if len(args) > 0 {
+			options.target = strings.Split(args[0], "+")
```
bjaglin: I guess that would mean we drop support for the implicit/default group

bjaglin:
Spontaneously, since `cmd`s semantics are tightly linked to containers, and because the whole point of `crane` is to deal with several containers at the time, I would be more tempted to go another way if I had to retrofit `cmd` into the CLI, namely to keep potential commands "close" to their relevant containers: `crane run -dall foo='cmd for foo' bar='cmd for bar' ` to run foo and bar dependencies, & foo/bar with their respective overridden/explicit `cmd`. Granted, that's a pretty clunky syntax, especially when you start only one container, but I guess you get my point :) Or maybe a new command that has different semantics for the arguments could be introduced, since I am assuming 99% of the time, people will indeed override `cmd` for a single container?

I will try to give more thoughts to https://github.com/michaelsauter/crane/issues/195 (which I had missed) - there is definitely some work needed to accommodate for more uses cases without compromising on ease of use for the basic flows.

michaelsauter:
Yes, this is definitely just a hack and not intended to go into Crane as-is. More a starting point for a discussion :)

I agree that most people are likely going to use this for single containers. This would also be my use case. However, I don't see why this couldn't make sense for multiple containers too.

You write that Crane's whole point is to deal with several containers, but I slightly disagree. I think it's also useful if there is just one container. It simply removes the need to specify the correct run arguments. It is a lot easier for an end user to say `crane run container` than to read some documentation which tells them what arguments to pass to `docker run`.

That said, here's my main use case for this:

I have a Rails container, which basically starts `rails s` as the process. Of course, devs often want to run the console, so they do `docker exec -it <container> rails c`. That works, but it is not perfect:

a) you need to have the server running in order to get into a console
b) if the server would start in the foreground and people would close it, they couldn't start a console immediately (which is weird for them coming from a non-Docker workflow)
c) if the server starts in the background, it is complicated/odd to see the output (`docker logs -f <container>`), and even then you can't use  something like pry (for that you have to `docker attach`, which is weird since when you `ctrl+c` out of it you end the process)

Because of this, I would like to separate running a rails server and a rails console. Now one solution would be to have another container for executing rails c or tests or whatever, but I find that odd to have a running container which does nothing. The other option is to not use `docker exec` but to create a new container and remove it every time you want to run one command. For that to work, the command needs to be configurable. And to make the whole process smooth, it needs to fit well into the CLI. This is the reason I want to make `crane run <container> rails c` work. BTW, I think that docker-compose allows this.

I don't see any harm that this is most useful for single containers.

crane/cmd.go
michaelsauter: I don't want to drop the default group, I use that a lot. I might be missing something very obvious, but why does this prevent a default group? I guess it only prevents you from running a command against the default group without explicitly typing `default`. But you could still say `crane run`.

bjaglin: Right, I went a bit quick there... It's just the passing of commands to the default group which is slightly verbose, but as discussed above, it's probably not a very common use case.

michaelsauter:
BTW, alongside this, we'd probably need another change: allowing the containers to not be named (sth like a `name: false` or so). That way, you can have multiple containers running different commands based on the same crane config. Of course if you don't give a name to a container, there's no way for crane to address it later (e.g. to stop it). So either you'd want to set `rm: true` (which I think makes sense in most cases) or we'd do something with labels ...

dreamcat4:
Would not it be best to just perturb the container name with the cmd being run in that instance?

e.g. If the container is named `red is` --> `redis` + `-cmd-arg1-arg2` --> makes a container named: `redis-cmd-arg1-arg2`

dreamcat4:
Oh. It may be that is not possible if 2 instances with the same cmd+args. Then the container name would still conflict. Yet it could be perturbed in other ways, e.g. by appending the first 8 or 12 digits of the docker `container_id` field. e.g. `redis-5f9d33723ae5` or something like that (or a random one).

michaelsauter:
@dreamcat4 No need to complicate it, Docker already assigns random names :)

michaelsauter:
Implemented in #212.

Issue #199, opened 2015-06-11
Stop hooks not executed on kill / rm -f


michaelsauter:
For example `post-stop` is only executed for `crane stop`, not for `crane kill`.

I see 2 options:

1) Defining a `post-kill`, which is executed for `crane kill`. Probably `post-kill` should run for `rm -f`, too?

2) Executing `post-stop` also for `kill` and for `rm -f` if the container is currently running.

Personally, I like 2) for simplicity and clearer semantics around `rm -f`.

Thoughts?

bjaglin:
:+1: for 2)

PR #200, opened 2015-06-15
Run stop hooks for kill and rm -f

michaelsauter:
Addresses #199.

### DIFF
--- a/crane/container.go
+++ b/crane/container.go
@@ -744,9 +744,11 @@ func (c *container) Start() {
 // Kill container
 func (c *container) Kill() {
 	if c.Running() {
+		executeHook(c.Hooks().PreStop())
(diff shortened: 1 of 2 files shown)


crane/container.go
```diff
@@ -788,11 +790,13 @@ func (c *container) Unpause() {
 // Remove container
 func (c *container) Rm(force bool) {
 	if c.Exists() {
-		if !force && c.Running() {
+		containerIsRunning := c.Running()
```
bjaglin: `containerWasRunning` maybe?

bjaglin:
Maybe it's worth adding a note about that behavior in the doc?

crane/container.go
michaelsauter: I thought that too, but at that time it is still running. I don't think there is a name for it that really fits :(

PR #201, opened 2015-06-28
optimization: don't provision the same Dockerfile/image more than once

bjaglin:
Use-case: a `crane.yml` containing many sidekick containers using the `busybox` image or many databases using the same image but different environment variables to be bootstrapped; when a `lift` command is issued, there is no need to pull the same image over and over again.


michaelsauter:
This is nice, thanks!

PR #208, opened 2015-07-07
WIP Kingpin

michaelsauter:
Very ugly at this stage, and non-functional. 

**EDIT: This is probably becoming Crane 2.0, see below.**


michaelsauter:
This is turning into something like a "2.0" branch :)

* Changed semantics of run / start (replace recreate flag)
* Single target only
* Automatic resolution and handling of dependencies
* Extension of target to affected containers and/or dependencies (replace current cascade flags)
* Randomly named containers
* [Still to come] Ad-hoc cmd for run
* [Still to come] Exclude groups/containers (replace ignore missing flag)

michaelsauter:
This should be functional now, plus it has added support for:

* Ad-hoc cmd for run
* Exclude groups/containers (remove need for ignore-missing flag)

Would be interesting to get some user feedback on this. Might need to release that as 2.0 beta, once there CI is green again.

utaal:
Happy to help test this. I'm going to try and depend on the new api for our internal CI project and report back.

michaelsauter:
@utaal Awesome! Looking forward to what you think.

michaelsauter:
Closed in favour of #212.

Issue #207, opened 2015-07-07
Change run / start etc. semantics


michaelsauter:
As stated in #195, I'm not entirely happy about the current semantics. Following is a quite drastic change of semantics, but so far, it sounds much better to me. I'm toying with the idea of implementing that for Crane 2.

The basic idea is the following:

* Commands apply to a target (any number of containers or groups)
* Crane does whatever it needs to do in order for the command to work. That means, it will satisfy dependencies automatically if it can.
* There are 2 "dynamic groups": `dependencies` and `affected`. Using them in the target means that the command will be applied to dependencies or affected containers too
* This removes the flags `--cascade-dependencies`, `--cascade-affected` and `--recreate` (see below)

Examples:

* `crane start web`: Creates the `web` container if it does not exist (!). Starts the `web` container if it exists. Does nothing if the `web` container is running. It will also ensure all dependencies of `web` are in their required state (running or existsing).
* `crane run web`: Removes the `web` container if it does exist (!), then runs it. It will also ensure all dependencies of `web` are in their required state (running or existing). Note that it does not remove any of the dependencies before running/starting them. 
* `crane kill web`: Kills the `web` container if it is running, otherwise it does nothing
* `crane lift web`: Provisions the `web` container, then the same as `crane run web`. Note that it does not provision or remove any dependencies before running/starting them.
* `crane run web+dependencies`: Same as `crane run web`, but it will also remove any dependencies before running them. E.g. if `web` has just one link to `postgres`, then `crane run web+dependencies` is the same as `crane run postgres && crane run web`

I understand this is quite different to how it works currently, but I believe it is much better. Some advantages:
* `--cascade-dependencies`, `--cascade-affected` are cumbersome to type or remember
* Maintaining groups to include all dependencies of a container is cumbersome, too
* `crane start` would not always work since a container might not exist yet, but always using `crane run` would lead to warnings that the container was not recreated. With the above proposal, you can either use the "safe" `crane start` or the "dangerous" `crane run` (which equals the current `crane run --recreate`). BTW, I'm unsure whether this violates the Crane principle of "map directly to a Docker command". However, we're already doing some smarts as to whether a command actually needs to run. Also, you could argue that actually running a container even if it exists already is mapping more closely. In the end, I think this is an acceptable trade-off.
* It is much easier to implement `crane run web rails s -p 0.0.0.0`, since that will bring up any required dependencies automatically
[...]

PR #212, opened 2015-07-21
Crane 2.0

michaelsauter:
Continued from #208.

Major changes:
- [x] Changed semantics of lift / run / start (remove need for recreate flag)
- [x] Single target only (a change needed for ad-hoc commands)
- [x] Automatic resolution and handling of dependencies
- [x] Extension of target to affected containers and/or dependencies (removes need for cascade flags)
- [x] Ad-hoc cmd for lift / run / create
- [x] Uniquely named containers (allowing multiple containers of the same template with different commands)
- [x] `--exclude` groups/containers (removes need for ignore-missing flag)
- [x] Trigger `post-start` hooks after `start` event is sent
- [x] Adds a `crane exec` command, which starts containers first if needed

Todo:

- [x] Update documentation
- [x] Wait for feedback
- [x] Decide whether to change `dockerfile` key into a  `build` object


michaelsauter:
Since this has diverged significantly from master, I have included the changes from master here since 1.5.0.

michaelsauter:
Todo:

[ ] Update documentation
[ ] Wait for feedback

bjaglin:
Didn't have the time to try it out yet, but I am obviously very interested to check whether it fits our use-cases. I should be able to get back to you with exhaustive feedback within 10 days!

michaelsauter:
Awesome, thanks!

I hope that it still allows to do what was possible previously, but be more flexible. So for example, you can still work with groups like you could before, but the automatic cascading probably removes some use-cases for groups.

One thing that might affect you IIRC is that there is only one target now. This is due to the fact that you can specify ad-hoc commands now. I think it also serves as a simplification, which is nice. If we do need more powerful targets, we could think about groups in groups again :)

bjaglin:
In the meantime, since you seem pretty confident about the new direction, what about officially releasing a 2.0-rc1 from a 2.0 branch that would still allow breaking changes? I guess that's how you can get more feedback and bugs reported.

Also, I see this is targeting master - can we branch off 1.x to do maintenance/docker flag updates there as well?

michaelsauter:
Sounds like a plan! Will do once I get to it.

michaelsauter:
Binaries can be found at https://github.com/michaelsauter/crane/releases/tag/v2.0.0-rc1.

bjaglin:
Thanks, that would make it easier to get feedback! FYI, I marked it as a "pre-release" in the release page.

bjaglin:
How should we handle the feedback? tickets/PRs onto that branch for bugs? questions here about semantics?

bjaglin:
graph seems to be broken
```
brice@t440s:~/sandbox$ cat crane.yml 
containers:
  hello:
    image: busybox
    run:
      cmd: echo hello world
brice@t440s:~/sandbox$ crane2 graph
digraph {
}
brice@t440s:~/sandbox$ crane2 graph hello
digraph {
}
```

bjaglin:
missing depencies are handled gracefully
```
brice@t440s:~/sandbox$ cat crane.yml 
containers:
  hello:
    image: busybox
    run:
      cmd: echo hello world
      link:
        - foo:foo

brice@t440s:~/sandbox$ crane2 run hello
ERROR: runtime error: invalid memory address or nil pointer dereference
```

bjaglin:
shouldn't ensuring a `volumes-from` dependency exists be implicit?
```
brice@t440s:~/sandbox$ cat crane.yml 
containers:
  hello:
    image: busybox
    run:
      cmd: echo hello world
      volumes-from: 
        - volumes-from

  volumes-from:
    image: busybox
    detached: true
    run:
      cmd: echo volumes-from
brice@t440s:~/sandbox$ crane2 run hello
Removing container hello ... hello
Running container hello ... Error response from daemon: no such id: volumes-from
ERROR: exit status 1
```

michaelsauter:
Yikes! Thanks for the extensive feedback. Will need to look into those.

Best would be to open any PRs into this branch I think.

michaelsauter:
@bjaglin `graph` command should work again now. Didn't realize it was using the modified version of the dependency graph, which in your case had no containers left anymore ...

michaelsauter:
Also fixed the 2 other issues now.

Probably need to look into how Crane behaves for unique containers in those situations. That is probably still very sketchy.

crane/unit_of_work.go
```diff
@@ -0,0 +1,239 @@
...
+	c := []Container{}
+	for _, name := range uow.order {
+		if includes(uow.targeted, name) {
+			c = append(c, cfg.Container(name))
+		}
+	}
+	return c
+}
+
+func (uow *UnitOfWork) Affected() []string {
```
bjaglin: `Affected()` has a completely different meaning when it comes to targeting/cascading. Shouldn't it be `Required()` or `Dependencies()`?

crane/target.go
```diff
@@ -0,0 +1,139 @@
...
+	targetParts := strings.Split(targetFlag, "+")
+	targetName := targetParts[0]
+	extendDependencies := false
+	extendAffected := false
+	for _, v := range targetParts[1:] {
+		if v == "dependencies" || v == "d" {
+			extendDependencies = true
+		} else if v == "affected" || v == "a" {
+			extendAffected = true
+		} else {
```
bjaglin: we should be more verbose when facing unknown qualifier(s)

crane/target_test.go
```diff
@@ -0,0 +1,83 @@
...
+		{
+			target: "b+affected",
+			expected: Target{
+				initial:      []string{"b"},
+				dependencies: []string{},
+				affected:     []string{"a"},
+			},
+		},
+		{
+			target: "b+dependencies+affected",
```
bjaglin: test the short version of the cascading flags (since they are documented)

crane/container.go
```diff
@@ -488,61 +535,63 @@ func (c *container) Provision(nocache bool) {
...
+	// Detach
+	if c.RunParams.Detach {
+		args = append(args, "--detach")
+	}
+	args = append(args, c.createArgs(cmds, excluded, configPath)...)
+	c.executeArgsWithStartEventObserver(args)
+}
+
+func (c *container) executeArgsWithStartEventObserver(args []string) {
+	cmd, cmdOut, _ := executeCommandBackground("docker", []string{"events", "--filter", "event=start", "--filter", "container=" + c.ActualName()})
```
bjaglin: no need to run this if there is no hook to execute later

bjaglin:
I finally found some time to play around with that branch. Great work overall, even though the UX changed significantly, I was able to adapt rather fast. 2 things come to my mind:
* The `unique` name feature is cluttering the code and doesn't seem to fit with the rest of the UX at all. Maybe that can be worked on in a further iteration?
* The same way dependencies are ensured to be started when a container is created/started/lifted/etc, I was expecting pause/stop/remove commands to be propagated to the affected containers, since these will end up in an inconsistent state. I guess this is the same debate we had a few months ago when I was suggesting to make `--cascade-affected=all` a default. This can result in unexpected state loss, but an interactive question could be raised. The behavior of the symmetrical command after issueing a "negative" order is also questionable, as the user would need to do a `unpause +affected` after a `pause` to get back to the previous state.

crane/unit_of_work.go
michaelsauter: I am not sure what you mean - affected still behaves very much like before, no?

crane/target.go
michaelsauter: Right, will have to add that.

crane/container.go
michaelsauter: Good point, will add as well.

michaelsauter:
Thanks for your feedback!

Regarding `unique`: Basically, my main motivation to do Crane 2 was to introduce ad-hoc commands, since I think that's really handy in dev and CI environments, where I see Crane used most. In order to do ad-hoc commands properly, I think `unique` is required. So - I really want to get that into Crane 2 :) What exactly did you find difficult / unexpected about it? I thought it worked quite well when I used it so far.

Regarding `affected`. It seems to me that you misunderstood how it actually works. First off, my idea of what Crane does in version 2:
* Apply the specified command to the given target
* Do everything necessary to achieve this (e.g. start dependencies)

I think that is a good way of doing things. Besides, if you continue to use groups, it basically still works exactly the same like in version 1 (given you did not use the cascade flags). Version 2 just allows you to use groups less (which again, is a requirement for ad-hoc commands in my eyes).

I do not really understand your pause / unpause example ...Say you have container A depending on container B which depends on C. If you run `crane pause A`, it pauses A and nothing else, so symmetrically you can say `crane unpause A`. The same would be the case if you targeted B or C. If you said `crane pause A+dependencies`, then you would have to do a `crane unpause A+dependencies` to revert it (or alternatively, `crane unpause C+affected`).

And yes, if you for example kill a container which is a dependency of another container, then that container is broken. That was the case in version 1, and it is the case in version 2. Also, it is what happens in plain Docker, so I think that is okay. I would be happy to output some information to the user to show potential problems and solutions (e.g. use `+affected`), but I don't think it should be the default. By default, Crane commands should be safe I think.

Does that clear up your questions?

michaelsauter:
@bjaglin Addressed your (code) feedback.

crane/unit_of_work.go
bjaglin: Affected" in the cascading context means ""containers that depends directly or indirectly on the given target". From what I see here, it's more "containers that a given target depends on, directly or indirectly".

bjaglin:
:+1: for code changes, will answer back regarding the rest soon.

bjaglin:
About the affected discussion: say `a` depends on `b` and they are both started. `docker pause b` will potentially turn `a` into an unstable state as one of its dependency doesn't answer. We have three choices when it comes to expected behavior:

1. don't do anything - that's crane 1.x behavior
2. warn the user that affected containers (listing them out) might be impacted by this command
3. cascade the command issued to the target to the affected (making `+affected` the default)

Even though I understand 1. is the docker behavior, I don't understand why 3. is "unsafe". After all, crane is an orchestration tool allowing you to reach a desired state, and a broken state is rarely something you want to reach. That's what I meant by "side effects" in https://github.com/michaelsauter/crane/issues/207#issuecomment-120625368. But I'd be happy to implement 2. in a separate PR if you don't believe in 3.

README.md
```diff
@@ -227,11 +241,22 @@ Hooks can be defined on a group level (`foo`, `bar`) so that they apply to all c
...
+* `post-build`: Executed after building an image
 * `pre-start`: Executed before starting or running a container
 * `post-start`: Executed after starting or running a container
 * `pre-stop`: Executed before stopping, killing or removing a running container
 * `post-stop`: Executed after stopping, killing or removing a running container
 
+Every hook will have the name of the container for which this hook runs available as the environment variable `CRANE_HOOKED_CONTAINER`.
+
+### Container Prefixes
+It is possible to prefix containers with a global `--prefix` flag, which is just
```
bjaglin: don't wrap lines?

bjaglin:
Regarding the unique feature: apart from the implementation a bit being "messy" because a lot of the code paths need to be blocked when that flag is set (and there is no way around it given the current architecture, I am not criticizing the style nor advocating for a refactor), I find the UX to be awkward:
* I first interpreted "unique" as "singleton", but it's the exact opposite
* What's the point of giving a random name at all?
* How does/can uniquely-named container work as link targets?
* Labels could be used to identify uniquely-names container and this support most of the commands, allowing to control several instances of the same container definition via a single command

I actually find the pre-https://github.com/michaelsauter/crane/commit/fc4671153fa25f5661a77af81ec41a8e68dde23f design better. I must say I don't have a better proposal at the moment, but I am worried releasing a 2.0 like that will set that feature in stone. Maybe tagging this as experimental would be a good tradeoff to iterate?

In any case, that's not a feature we are going to use, so I don't consider this a showstopper for releasing 2.0, I am just giving my 2 cents :)

bjaglin:
Maybe time for a rc2?

crane/unit_of_work.go
michaelsauter: I'm going to have a second look at the code :) I thought and still think it does exactly what you expect (extending "upwards").

I definitely want the semantics like in Crane 1.

README.md
michaelsauter: Why not? :)

I personally prefer to wrap the lines since that makes the source file more readable and doesn't affect the display.

michaelsauter:
Thanks again for the extensive feedback, and yes, time for rc2 once I get to it!

Regarding `unique`:

* If you have a better name for it, I would be totally up for changing it
* The main use case is this: I would like a general purpose container that I can run multiple commands for. So for example, I have a Rails container called `web`, that allows me to do `crane run web rails` and `crane run web bash` (and whatever else I want to do) at the same time. If the container had the same name, I could only run one at the same time, which is not really an option. Consider running tests using that container (which might take a long time) - you would not want to wait until they finish to try something in the console.
* So basically, `unique` only really make sense for containers "at the top level". Crane does work when a linking to a container marked as `unique`, but I don't see a good use case for that
* I thought about restricting `unique` a little, by only allowing it to be used when `rm` is set to `true` as well ... since that is most likely what you want, or else you end up with endless containers
* Using labels to reach them later is definitely a good idea and something I would like to do, but don't think this needs to be added now (since it would only improve Crane, not break existing behaviour)
* Marking this as experimental sounds like a good idea initially, I will think about it.

Apart from that, does the UI look good for your use-cases? Would be interested whether you benefit from Crane 2 or find it more difficult to work with.

crane/unit_of_work.go
bjaglin: My comment was jut about code naming - `affected` in the CLI does behave the same, it's just that the function here is called `Affected()` but actually relies on https://github.com/michaelsauter/crane/pull/212/files#diff-e6ff2f0b5db638b571478fa48f4676d4R33 and therefore goes "downwards" AFAICS.

README.md
bjaglin: no strong opinion, I just noticed it because I was on a small screen with the "side by side" view :)

crane/unit_of_work.go
michaelsauter: Finally I got you :) 

Makes a lot of sense, I couldn't remember my own code. I'm going to rename this.

crane/cmd.go
```diff
@@ -2,298 +2,313 @@ package crane
...
+func handleCmd() {
+	switch kingpin.MustParse(app.Parse(os.Args[1:])) {
+
+	case liftCommand.FullCommand():
+		commandAction(*liftTargetArg, func(uow *UnitOfWork) {
+			uow.Lift(*liftCmdArg, excluded, *liftNoCacheFlag)
+		}, true)
+
+	case versionCommand.FullCommand():
+		fmt.Println("v2.0.0-rc1")
```
bjaglin: @michaelsauter looks like your forgot to bump that for rc2

michaelsauter: Ah thanks. Guess the release script does not handle that properly ...

bjaglin:
Thanks for the details about `unique` - I guess I will have more feedback if/when we start using it, but as I said, LGTM now.

Regarding the UI, `--exclude/-e` is actually the most promising flag (so that a dev can start all dependencies of `foo`, but without `foo` itself for example). However, it still seems broken ATM (rc1 was panicking, rc2 is giving a weird message). In the lack of a test/PR:

```
brice@t440s:~$ cat crane.yml 
containers:
  hello:
    image: busybox
    run:
      link: ["sleep:sleep"]
      cmd: echo hello 
  sleep:
    image: busybox
    run:
      detach: true
      cmd: sleep 10
brice@t440s:~$ crane2 run 
Removing container sleep ... sleep
Running container sleep ... 4d2101f3c604759fffb2a3cc93e3dac26f04e55dda65ca4622a3f89cb7979493
Removing container hello ... hello
Running container hello ... hello
brice@t440s:~$ crane2 run -e sleep
ERROR: Container sleep referenced, but not defined.
brice@t440s:~$ crane2 run -e hello
ERROR: Container hello referenced, but not defined.
```

For the rest, it's mostly good, but I haven't found a way to lift more than one component and their dependencies (an extremely common use case for integration tests) without restarting their common dependencies (if run had an opt-in `-r/--restart` flag, I could just issue n commands in a row, but that's not the case). I am thinking about extending the target syntax to support multiple containers `foo+bar+qux`, but I don't have much time right now, so we'll hold the upgrade to 2.x until this is implemented, we are not in a hurry anyway.

bjaglin:
Oh and the other showstopper for the upgrade would be the lack of message when doing a destructive action (last paragraph of https://github.com/michaelsauter/crane/pull/212#issuecomment-129040300) - people are currently used to do `-d all` and `-a all` with Crane 1.x but because the `-d all` is becoming implicit (at least "associated" are always included, so for 99% of the users it's the same), I am expecting a lot of tears in the QA team when they realize someone (that thinks cascading flags are now useless) restarted a backend without its affected containers, breaking frontends because the backend IP changed :/

michaelsauter:
Regarding lifting more than one component without restarting their common dependencies:

Say you have container A and B, which have some common dependencies. If you now have a group G which consists of A and B, when you say `crane lift G`, it should give you what you want. Or do you want to do something different?

Your second showstopper is interesting, because I think how Crane works in v2 is consistent, but somehow I haven't been able to communicate it clearly so there is some misunderstanding. I have to make sure I get that right before the release.

There are basically 2 rules:
1) Crane will apply the command ONLY to the target
2) Crane will od with other ("associated") containers what it has to do in order for (1) to succeed

That means that `-d all` is NOT becoming implicit. It still is explicitly needed. If you have container A with a link to container B, then if you execute `crane run A`, then `A` will be recreated, but `B` will not. If you want that, you will need to say `crane run A+d`.

Because of this, I don't think that the behaviour changes at all for your use case. We still need better messaging though, 100% agree. I have thought more about making affected etc. the default behaviour but I just can't get that to work out.

michaelsauter:
@bjaglin Also, I would like to keep having just one target on the CLI. If it turns out we need multiple groups after all, I'm more in favor of putting this into the configuration (by allowing groups inside groups). I was against this in the past, but now I think it's preferable to multiple targets, especially now that we have those "target extensions" `dependencies` and `affected`. How multiple targets would be treated if specified on the CLI would be a bit unclear to me (e.g. does the order affect it?). I think it is more obvious when done in the configuration.

PR #214, opened 2015-07-22
Experiment with overriding config

michaelsauter:
Allows users to override any values in the Crane config with a `crane.local.yaml`.
The local file uses the same structure as the normal config, and just overrides whatever value has been set for a certain key.

Thoughts?

### DIFF
--- a/crane/config.go
+++ b/crane/config.go
@@ -80,14 +80,14 @@ func findConfig(location string) string {
 
 // readConfig will read the config file
 // and return the created config.
-func readConfig(filename string) *config {
+func (c *config) readFile(filename string) {
 	data, err := ioutil.ReadFile(filename)
 	if err != nil {
 		panic(StatusError{err, 74})
 	}
 
 	ext := filepath.Ext(filename)
-	return unmarshal(data, ext)
+	c.unmarshal(data, ext)
 }
 
 // displaySyntaxError will display more information
@@ -116,13 +116,12 @@ func displaySyntaxError(data []byte, syntaxError error) (err error) {
 
 // unmarshal converts either JSON
 // or YAML into a config object.
-func unmarshal(data []byte, ext string) *config {
-	var config *config
+func (c *config) unmarshal(data []byte, ext string) {
 	var err error
 	if ext == ".json" {
-		err = json.Unmarshal(data, &config)
+		err = json.Unmarshal(data, &c)
 	} else if ext == ".yml" || ext == ".yaml" {
-		err = yaml.Unmarshal(data, &config)
+		err = yaml.Unmarshal(data, &c)
 	} else {
 		panic(StatusError{errors.New("Unrecognized file extension"), 65})
 	}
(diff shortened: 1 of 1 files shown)


bjaglin:
Interesting - I can definitely see a use-case for it! I didn't have the chance to play with it - how does it  work when overriding a "nested" key (say the publish port of a container declared in the main config) ?

michaelsauter:
It seems to just "overlay" the second config over the first one. So if the second one is:
```
containers:
  foo:
    run:
      detach: true
      publish: ["3000:3000"]
```

And the second one is:
```
containers:
  foo:
    run:
      publish: ["8080:8080"]
```

Then it is treated as if it had been:
```
containers:
  foo:
    run:
      detach: true
      publish: ["8080:8080"]
```

bjaglin:
Nice :+1: However, I am not sold on the name (what's local about it? override sounds more clear to me) nor, to a smaller extend, the semantics. How about tweaking the tree traversal up to `/` to support several `config.yml`, the ones closer to the path overriding values of the furthest ones?

michaelsauter:
`override` sounds good to me too. Let's use that.

I thought about the tree traversal thing too, and in general like the idea. However, sometimes (often?) you would need the overriding config in the same directory for it to make sense.

One example use case is this:
Say I have a Rails project that defines a `crane.yml` at the root. Now I want to test the project with a new Ruby version, so I use a `crane.override.yml` in the root to change the image tag, since I don't want to force everyone into upgrading just yet. I can't place the overriding crane config anywhere else than on the same hierarchy level of the normal config.

dreamcat4:
Was more hoping for some form of manual include directive instead of this. Just for (my own) need to split up 1 very large `crane.yml` file into multiple smaller files.

bjaglin:
All right, I see your use-case @michaelsauter . Maybe `-c/--config` could be repeated/take an (ordered) list to cover @dreamcat4 use-case too?

dreamcat4:
Ah to let multiple config file switches on cmdline is a bit different than the 'include directive' (which is more like the 'C' `#include` directive). Sorry there are overlapping features. I'm not clear which is best for which use-case. I just stated my own use-case. So perhaps you can know if there is better way to solve it.

FYI:

Now also remembering something else about this... It may be useful to consider employing a globbing / regex pattern. For example `**/*.myfiles.yml` to pickup / dynamically select the matching files for processing. That is a feature which requires something like the unix `find` tool. Or whatever the equivalent is of that for the 'go' filesystem APIs. Depending how well 'go' supports such globbing functions already (the pre-existing builtin support).

michaelsauter:
I think we're talking about different use cases here:

1) overriding some configuration locally (e.g. because the branch you're working on needs a different image)

2) managing multiple crane files and merging them together (include directive)

I see them as orthogonal. I know people want (2), and I'm not against it. However I don't have a need for it right now so I won't contribute it :) I do have a need for (1) though, so this is probably going to be merged soonish (with slight changes compared to this PR + tests).

michaelsauter:
Closing this for now. Might re-open with different implementation when there is a need for it (right now there is not).

PR #215, opened 2015-07-28
Allow running and lifting individual containers using a custom name.

jesper:
Not entirely sure if this makes sense for merging upstream, but maybe?

We have a use-case where we need to be able to occasionally launch an extra instance of a container on demand. This patch adds at "--custom-name" parameter that can be used if a single container target has been supplied to a "crane lift" or "crane run".

Caveat: These containers become "orphans" - crane cannot keep track of them, and as such, you'll need to manage that yourself. I have a few ideas about how this could be integrated in crane (Labels?) - but I'm guessing it's out of scope for the project, and have therefore left it out.

### DIFF
--- a/crane/cmd.go
+++ b/crane/cmd.go
@@ -24,6 +24,7 @@ type Options struct {
 	ignoreMissing       string
 	config              string
 	target              []string
+	customName          string
 }
 
 var options Options
@@ -66,7 +67,7 @@ func handleCmd() {
 		Long: `
 lift will provision missing images and run all targeted containers.`,
 		Run: configCommand(func(config Config) {
-			config.TargetedContainers().lift(options.recreate, options.nocache, options.ignoreMissing, config.Path())
+			config.TargetedContainers().lift(options.recreate, options.nocache, options.ignoreMissing, config.Path(), options.customName)
 		}, false),
 	}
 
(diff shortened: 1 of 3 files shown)


michaelsauter:
Not sure if you've seen this, but I have a similar use case as well. See #212 for the work in progress.

I have a binary built from that branch over the last week or so and there were no problems so far. So this might become 2.0 relatively soon.

It does not allow you to specify a custom name on demand, but it does allow you to define a container as "unique", which will add  a timestamp to the container name. It's also limited in that Crane can't keep track of them (so I would envision people using `unique: true` and `rm: true` in tandem).

Let me know if that solves your issue. Otherwise, have you looked into whether `docker exec` will provide what you need?

michaelsauter:
(That said, your suggestions could still be relevant in Crane 2, but I'm hesitant to merge it before then)

jesper:
Thanks for pointing out that feature in #212. I think it may do the trick for our use-case (when combined with labels).

dreamcat4:
Hmm. For cleaning up errant (orphaned) containers, the Sherdock tool may be useful

https://github.com/rancher/sherdock

It launches a web 2.0 interface for administration. This fragment can be put in `crane.yml` to try it out:
```yaml
containers:
  sherdock:
    image: rancher/sherdock
    run:
      volume:
        - /var/lib/docker:/var/lib/docker
        - /var/run/docker.sock:/var/run/docker.sock
      publish:
        - 8080:8080
      detach: true
```

michaelsauter:
I'm going to close this since both unique containers and container prefixes should provide enough flexibility to achieve what you have in mind. Hope that's fine with you.

PR #217, opened 2015-07-30
Proposal: build hooks

t-suwa:
Crane does not support build hooks currently. But I want to execute some extra scripts before/after building an image. For example:

- get latest sources for the `<src>` of Dockerfile's ADD/COPY instruction 
- asset pipeline
- talk to bots

I think build hooks make the provision more flexible.


michaelsauter:
Looks great, thank you! I might hold off merging this until 2.0 is released (which is hopefully soon). It might need rebasing then, but should be very little if any.

t-suwa:
Wow thank you!  I'm glad since this is my first PR.

michaelsauter:
Actually, would you mind rebasing that on top of the 2.0 branch and open a new PR against that? Then we could include this in 2.0.

t-suwa:
I see. I'm going to open a new PR based on the 2.0 branch as soon as possible.

michaelsauter:
Merged into 2.0 now.

PR #219, opened 2015-08-06
ensure multiple references to the same container is handled

bjaglin:
I could not execute any command because it turned out that I had the same container referenced several times in a group definition - and this is working fine with Crane 1.x.


michaelsauter:
Good catch. Thanks!

PR #223, opened 2015-08-25
Add `--tag` global flag

t-suwa:
Using '--tag' global flag, we can easily handle containers with distinct names.

- `crane provision --tag latest`
- `crane lift --tag 2.0.0-rc2 --prefix rc2`

But crane 2.0 is about to be released, so you are welcome to skip this PR.


crane/config.go
```diff
@@ -219,6 +227,20 @@ func (c *config) initialize() {
...
+	// At the moment, only `--tag` flag triggers override
+	if len(c.Tag()) == 0 {
+		return
+	}
+
+	for _, container := range c.containerMap {
+		container.OverrideImageTag(c.Tag())
+	}
+
+	os.Setenv("CRANE_TAG_FLAG", c.Tag())
```
michaelsauter: I think we should not set this here. Instead, we can have kingpin read the value of `tag` from `CRANE_TAG`, which can then be used in hooks too. What do you think?

michaelsauter:
Thanks for this!

Looks good generally, and I can see how this would be useful (especially together with prefixing). I think we can look at this after the 2.0 release. Once that is out, can you adjust the readme and rebase?

crane/config.go
t-suwa: Thaｔ’s a good idea!  I totally agree with you. Maybe I can make an update within the next week.

t-suwa:
I'd love to. ;-)

t-suwa:
This PR was rivised to #229.

PR #224, opened 2015-08-25
Add CRANE_PREFIX environment variable support.

tmc:


crane/cmd.go
```diff
@@ -2,9 +2,10 @@ package crane
 
 import (
 	"fmt"
-	"github.com/alecthomas/kingpin"
 	"os"
 	"strings"
+
+	"github.com/alecthomas/kingpin"
```
michaelsauter: Did you run that through gofmt? I'd expect no blank line here ...

michaelsauter:
I like it! Good idea. Will merge once the formatting issue is addressed.

bjaglin:
:+1: should we generalize that to other CLI flags for consistency? `CRANE_VERBOSE`, `CRANE_CONFIG`

crane/cmd.go
tmc: that's the output from goimports - https://godoc.org/golang.org/x/tools/cmd/goimports

michaelsauter: Ah okay - then that's all good. I wanted to get that running a while back but for unclear reasons it did not work ... have to try that again, looks simple to setup.

michaelsauter:
@bjaglin Sounds good to me. @tmc What do you think? Want to add them as well?

tmc:
@bjaglin @michaelsauter yeah I think for consistency they should also be added.

PR #226, opened 2015-08-25
Additional tag support

tmc:
I realize there's hesitation around enhancing crane's build features but this is a simple implementation of expanded tagging capabilities.

```yaml
containers:
    base:
        image: e_corp/base
        build:
            context: dockerfiles/e_corp/base
            tags:
             - ${IMAGE_TAG}
    go1.5:
        image: e_corp/go1.5
        build:
            context: dockerfiles/e_corp/go1.5
            tags:
             - ${IMAGE_TAG}
```
```sh
IMAGE_TAG=$(git rev-parse --short HEAD) crane provision
```
This is my use case -- e_corp/go1.5/Dockerfile uses a "FROM e_corp/base".


michaelsauter:
Hmm, not too bad :) I like how transforming the `dockerfile` key into a `build` map helps those things. 

Similar to your other pull request, I would like to address this after 2.0.

michaelsauter:
Could you rebase this to contain just the tag support?

michaelsauter:
I thought a little about this and concluded that `tags` is not a good fit since it doesn't have an equivalent in plain Docker. However, I am open to add just one `tag`, since that exists in Docker. What do you think?

I am not sure what the interaction would be with the global `--tag` flag. Should it override this here or not?

tmc:
I really do want more than one tag as I want layered images to be built in a CI context and need :latest to be tagged for descendant images and :$(tag) to be tagged for running a specific set of images.

tmc:
closing and re-opening as the PR should be agains master (since 2.0 has merged in since creation)

tmc:
closing in favor of #239

PR #232, opened 2015-10-13
Add `protected` key to keep data volume containers (experimental)

t-suwa:
By using data volume containers, we can remove and create application containers repeatedly without risk of data loss. 

- [Managing data in containers](http://docs.docker.com/userguide/dockervolumes/)
- [Why Docker Data Containers are Good](https://medium.com/@ramangupta/why-docker-data-containers-are-good-589b3c6c749e)

Unfortunately, Crane always removes the existing containers when invoking `lift` or `run`. This makes it difficult to use data volume containers.

So this PR introduces new `protected` key to keep data volume containers, but the key name is obscure and needed to find more suitable one maybe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ The map of containers consists of the name of the container mapped to the contai
 
 * `image` (string, required): Name of the image to build/pull
 * `unique` (boolean, optional) `true` assigns a unique name to the container (experimental)
+* `protected` (boolean, optional) `true` prevent the container from being deleted (experimental)
 * `run` (object, optional): Parameters mapped to Docker's `run` & `create`.
 	* `add-host` (array) Add custom host-to-IP mappings.
 	* `blkio-weight` (integer) Need Docker >= 1.7
--- a/crane/container.go
+++ b/crane/container.go
@@ -23,6 +23,7 @@ type Container interface {
 	Exists() bool
 	Running() bool
 	Unique() bool
+	Protected() bool
 	Paused() bool
 	ImageExists() bool
 	Status() []string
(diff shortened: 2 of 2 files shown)


michaelsauter:
@t-suwa Can you post an example config? 

Crane leaves volumes as is without deleting them if you don't target them.

t-suwa:
OK, below is a typical use case of `protected` container.

```
containers:
  data:
    protected: true
    image: busybox
    run:
      volume:
       - /var/data
      detach: true
  app:
    image: nginx
    run:
      volumes-from:
        - data
      detach: true
```

Following command sequence shows my assumption:

```
$ crane lift # 1) first lift: create both data and app container
...
$ docker exec -it app ls /var/data # 2) no data found
$ docker exec -it app touch /var/data/foo.txt # 3) create a new file
$ crane lift # 4) second lift: delete & create app container (data container is untouched)
...
$ docker exec -it app ls /var/data # 5) data container is keeping same data
foo.txt
```

I see container group can be used for this purpose. However, IMHO, `protected` (or `persistent` ?) would be more intuitive and easy to use.

What do you think?

michaelsauter:
`crane lift app` should do exactly what you want. No need for extra config. Crane will take care of creating the volume the first time, then leave it as it is.

t-suwa:
I see your point. But if you forget to specify an `app` target accidentally, your data will become `dangling volume` and difficult to rescue.

It would be safer to specify  the `protected`  in crane.yaml only once instead of typing commands carefully.

Or, other alternatives?

michaelsauter:
Hmm. I see. Not sure how the upcoming Docker volumes API might change this situation. I think at the moment, I would like to see how that plays out, and then decide on this addition.

t-suwa:
Oh I see. Thank you for your time.

For the time being, I'll try to use container groups to protect data volume containers. 

Thanks anyway.

Issue #233, opened 2015-10-15
Crane doesn't support dns-search in run


scornelissen85:
I'm using docker run --dns_search. I think this option is missing in crane (or am I missing it somewhere?)

michaelsauter:
No I don't think that is currently supported. However, it should be really trivial to add. Do you want to give it a shot?

PR #234, opened 2015-10-15
Added support for dns_search in run

scornelissen85:
This should be a fix for #233. It add dns_search parameter in run and will pass this parameter to docker run.


michaelsauter:
Thanks for this. There are some issues here (e.g. wrong flag name) and since I was about to release the next version, I decided to quickly implement it in 77eead6.

PR #235, opened 2015-11-07
Fix some issues detected by go vet/golint

pdalpra:
This PR fixes some issues that got detected by `go vet`/`golint`:

* Usage of `\n` in `fmt.Println`
* Acronyms in names not in full caps
* `return` in `else` blocks
* unused second values in `for` loops

`golint` also states that anything exported should be documented, but I'll to dig deeper in the codebase to provide helpful comments :)

### DIFF
--- a/crane/container.go
+++ b/crane/container.go
@@ -41,7 +41,7 @@ type ContainerInfo interface {
 	ActualName() string
 	Image() string
 	ImageWithTag() string
-	Id() string
+	ID() string
 	Dependencies() *Dependencies
 	Unique() bool
 	BuildParams() BuildParameters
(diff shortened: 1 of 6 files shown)


crane/container.go
```diff
@@ -69,15 +69,15 @@ type RunParameters struct {
 	RawCapAdd       []string    `json:"cap-add" yaml:"cap-add"`
 	RawCapDrop      []string    `json:"cap-drop" yaml:"cap-drop"`
 	RawCgroupParent string      `json:"cgroup-parent" yaml:"cgroup-parent"`
-	CpuPeriod       int         `json:"cpu-period" yaml:"cpu-period"`
```
michaelsauter: I'm not entirely sure, but I think I read at some point that gaoling prefers camel case here. I definitely think that it is easier to read then the alternativ versions. Mind removing all of those changes?

crane/container.go
```diff
@@ -494,32 +492,32 @@ func (r *RunParameters) Cmd() []string {
...
 
-func (c *container) Id() string {
+func (c *container) ID() string {
 	if len(c.id) == 0 && !c.Unique() {
 		// `docker inspect` works for both image and containers, make sure this is a
-		// container payload we get back, otherwise we might end up getting the Id
+		// container payload we get back, otherwise we might end up getting the ID
 		// of the image of the same name.
-		c.id = inspectString(c.ActualName(), "{{if .State}}{{.Id}}{{else}}{{end}}")
+		c.id = inspectString(c.ActualName(), "{{if .State}}{{.ID}}{{else}}{{end}}")
```
michaelsauter: I think this is actually `Id` in the returned JSON from Docker.

pdalpra: woops, my bad !

crane/container.go
pdalpra: This is actually a rule from `golint`, taken from : https://github.com/golang/go/wiki/CodeReviewComments#initialisms
But I can remove them if you prefer :)

michaelsauter: Oh. I must have confused it then with something else. I guess it's better to follow the guideline then, so leave it in.

Thanks for pointing me to the document.

pdalpra: My pleasure :)

crane/container.go
```diff
@@ -1024,8 +1021,8 @@ func (c *container) prefixedName() string {
 }
 
 // Return the image id of a tag, or an empty string if it doesn't exist
-func imageIdFromTag(tag string) string {
-	args := []string{"inspect", "--format={{.Id}}", tag}
+func imageIDFromTag(tag string) string {
+	args := []string{"inspect", "--format={{.ID}}", tag}
```
michaelsauter: This should stay `Id`, too :)

pdalpra: fixed !

pdalpra:
Everything should be okay now, just rebased my PR against master :)

michaelsauter:
Thank you!

PR #239, opened 2015-11-18
Additional tag support

tmc:
This is a simple implementation of expanded tagging capabilities. This makes crane more useful when building sets of related images.

```yaml
containers:
    base:
        image: e_corp/base
        build:
            context: dockerfiles/e_corp/base
            tags:
             - ${IMAGE_TAG}
    go1.5:
        image: e_corp/go1.5
        build:
            context: dockerfiles/e_corp/go1.5
            tags:
             - ${IMAGE_TAG}
```
```sh
IMAGE_TAG=$(git rev-parse --short HEAD) crane provision
```
This is my use case -- e_corp/go1.5/Dockerfile uses a "FROM e_corp/base" (which assumes the :latest tag).

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,8 +146,9 @@ The map of containers consists of the name of the container mapped to the contai
 	* `attach` (boolean)
 	* `interactive` (boolean)
 * `build` (object, optional): Parameters mapped to Docker's `build`.
-	* `context` (string)
-	* `file` (string)
+	* `context` (string) The path to the build context.
+	* `file` (string) Equivalent to --file specification.
+	* `tags` (array) Additional tags to apply when the container's image is built.
 * `exec` (object, optional): Parameters mapped to Docker's `exec`.
   * `interactive` (boolean)
   * `tty` (boolean)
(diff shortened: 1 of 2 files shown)


michaelsauter:
Your example only shows one tag, so I'm still a bit lost why you need multiple tags. What am I missing? :)

tmc:
It's two because it doesn't show the implicit :latest tag.

I'm building a tree of images -- "FROM e_corp/base" in a descendant image assumes :latest.

When I then decide to run all the generated images I use the ${IMAGE_TAG} (which in my case is a vcs-generated value).

michaelsauter:
Sorry for leaving this one open for so long.

I'm going to close it. Somehow the semantics feel unclear to me. We have `CRANE_TAG`, and this should allow to override the tag you're building (though I haven't checked if that is actually the case). Sure, this doesn't allow you to tag an image twice, but this just doesn't fit well into the current architecture of Crane.

As a side note, I think it is considered good practice to specify an exact tag when using `FROM` in a Dockerfile.

PR #240, opened 2015-12-05
bump to golang 1.5

bjaglin:
This is of course just for tests, but maybe it reflects the environment you use for release? It should at least :)


michaelsauter:
Haha yea. I'm still using 1.4.2 but guess that should be upgraded.

PR #241, opened 2015-12-05
Run flags walkthrough

bjaglin:
https://github.com/michaelsauter/crane/commit/306af735838f90608f5c4abe6e48fbf3bd8dfb10 turned out to be more challenging than the others, and undercovered a bug (see commit description)

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,7 @@ The map of containers consists of the name of the container mapped to the contai
 	* `mac-address` (string)
 	* `memory` (string)
 	* `memory-swap` (string)
+	* `memory-swappiness` (int) Need Docker >= 1.8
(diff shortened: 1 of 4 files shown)


michaelsauter:
Thanks for that and the fix to the net flag!

bjaglin:
Coincidence - the regression is actually hitting us as we need this feature now... Any chance you could cut out a release?

michaelsauter:
Will do tonight. In theory you could do it as well with the release script, but I'm also signing the releases at the moment. Not sure how to go about that. 

Will be 2.3.0 then.

michaelsauter:
@bjaglin Done: https://github.com/michaelsauter/crane/blob/master/CHANGELOG.md#230-2015-12-09

PR #242, opened 2015-12-06
pass `excluded` instead of relying on the global (cli) one

bjaglin:


crane/container.go
```diff
@@ -40,7 +40,7 @@ type ContainerInfo interface {
 	ActualName() string
 	Image() string
 	ID() string
-	Dependencies() *Dependencies
+	Dependencies(excluded []string) *Dependencies
```
michaelsauter: I wanted to have it like that, then decided against it since this makes it non-trivial for the generate feature to use `Dependencies`.

michaelsauter: I know the global is ugly, but I didn't find a better solution. Basically all methods can't have arguments if you want to use them easily from a template.

bjaglin: Oh of course, I didn't realize that!

bjaglin:
Closing this as it's obviously causing more troubles than benefits... I'll keep the design issues in mind though in case I get to work around that code!

PR #243, opened 2015-12-06
affected should only include existing containers

bjaglin:
Crane2.x changed the behavior of `affected`, and I am not sure that's intentional. See https://github.com/michaelsauter/crane/pull/77 for the same discussion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,7 +94,7 @@
   if you wanted to make sure both `web` and `database` are recreated, you can
   use the target `web` and extend it with `dependencies` like so:
   `crane run web+dependencies`. `affected` works the other way around, so if
-  you want to recreate `database` and all contaienrs that depend on it,
+  you want to recreate `database` and all containers that depend on it,
   execute `crane run database+affected`. You may also use the shortcuts `d` and
   `a`.
 
--- a/README.md
+++ b/README.md
@@ -257,12 +257,12 @@ in this example. If `default` were not specified, then `crane lift` would start
 It is also possible to extend the target to related containers. There are 2
 different "dynamic" groups, `affected` and `dependencies` (both have a short
 version `a` and `d`). In our example configuration above, when targeting the
-`postgres` container, the `blog` container would be considered to be "affected".
-When targeting the `blog` container, the `postgres` container would be
-considered as a "dependency". Therefore `crane run postgres+affected` will
-recreate both `postgres` and `blog`. Similarly, `crane run blog+dependencies`
-will recreate `blog` and `postgres`. It is possible to combine `affected` and
-`dependencies`.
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -257,12 +257,12 @@ in this example. If `default` were not specified, then `crane lift` would start
 It is also possible to extend the target to related containers. There are 2
 different "dynamic" groups, `affected` and `dependencies` (both have a short
 version `a` and `d`). In our example configuration above, when targeting the
-`postgres` container, the `blog` container would be considered to be "affected".
-When targeting the `blog` container, the `postgres` container would be
-considered as a "dependency". Therefore `crane run postgres+affected` will
-recreate both `postgres` and `blog`. Similarly, `crane run blog+dependencies`
-will recreate `blog` and `postgres`. It is possible to combine `affected` and
-`dependencies`.
+`postgres` container, if it had been started earlier, the `blog` container
```
bjaglin: really not a big fan of line wrapping, it makes a small update look like a big thing :/

CHANGELOG.md
```diff
@@ -94,7 +94,7 @@
   if you wanted to make sure both `web` and `database` are recreated, you can
   use the target `web` and extend it with `dependencies` like so:
   `crane run web+dependencies`. `affected` works the other way around, so if
-  you want to recreate `database` and all contaienrs that depend on it,
+  you want to recreate `database` and all containers that depend on it,
```
bjaglin: just a typo

README.md
michaelsauter: He right, that's a big downside of it.

michaelsauter:
I'm going to revisit "affected" when Crane supports the new networking feature. "Links" won't break then - so targeting affected containers is not really necessary anymore. Or do you use it for anything else?

bjaglin:
Well apart from the obvious problem of getting a different IP when restarting (which could still be a problem with Dockers' network v2 despite the dynamic resolution in case a service caches the very first DNS resolution), I think it's useful to invalidate any potential cache that an affected service could have, so spontaneously, I think there is still a need for it.

michaelsauter:
Okay, thanks for input. Not sure if Crane should provide a fix for this problem. Given the architecture Docker promotes, shouldn't services be able to deal with IP changes? I know that would be a tough decision but then again, affected is kinda difficult to understand, the semantics are tricky and the code is not the best too. Anyway, just putting these thoughts out there for now.

bjaglin:
Good point. However, in my opinion, a truly dynamic approach can't work without some sugar outside containers. For example, k8s pods have a local, fixed haproxy (or a set of iptables since 1.1) to handle the volatility of distributed services and avoid putting advanced service discovery within the containers. But it might not be within the scope of a simple tool such as Crane... Do you happen to know how Compose handles it?

Interesting question at least!

PR #244, opened 2015-12-09
Revert "docker 1.8.0: `docker run --memory-swappiness=42`"

bjaglin:
The current logic turns out to be broken - if no `memory-swappiness` value is provided in the configuration, the flag is passed with a `0` value (while the default is -1). Apart from the fact that we don't actually want to pass 0, this simply forbid Crane usage on pre-1.8.0 engines (like our CI slave...), so until we figure out how to handle a flag with a non-0 default value, I suggest we revert that.

My bad, I should be running in verbose mode and/or with Docker 1.6.0 to detect this kind of regression, or spend some time on setting up integration tests ;)

### DIFF
--- a/crane/container.go
+++ b/crane/container.go
@@ -103,7 +103,6 @@ type RunParameters struct {
 	RawMemory            string      `json:"memory" yaml:"memory"`
 	RawMemoryReservation string      `json:"memory-reservation" yaml:"memory-reservation"`
 	RawMemorySwap        string      `json:"memory-swap" yaml:"memory-swap"`
-	MemorySwappiness     int         `json:"memory-swappiness" yaml:"memory-swappiness"`
 	RawNet               string      `json:"net" yaml:"net"`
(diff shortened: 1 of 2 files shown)


crane/container.go
```diff
@@ -805,10 +804,6 @@ func (c *container) createArgs(cmds []string, excluded []string) []string {
 	if len(c.RunParams().MemorySwap()) > 0 {
 		args = append(args, "--memory-swap", c.RunParams().MemorySwap())
 	}
-	// MemorySwappiness
-	if c.RunParams().MemorySwappiness > -1 {
```
bjaglin: The valid range is 0-100, so we can't simply do `> 0` here since that means people won't be able to use that value... Any idea? I guess we need to wrap that into something similar to `OptBool`...

bjaglin:
A proper fix turned out to be easier than I thought... https://github.com/michaelsauter/crane/pull/245

PR #245, opened 2015-12-09
don't pass `--memory-swappinness 0` if parameter is omitted

bjaglin:
Fixes compatibility with engine<1.8.0 and wrongly-forced option on engine>=1.8.0.
See https://github.com/michaelsauter/crane/pull/244.

### DIFF
--- a/crane/container.go
+++ b/crane/container.go
@@ -103,7 +103,7 @@ type RunParameters struct {
 	RawMemory            string      `json:"memory" yaml:"memory"`
 	RawMemoryReservation string      `json:"memory-reservation" yaml:"memory-reservation"`
 	RawMemorySwap        string      `json:"memory-swap" yaml:"memory-swap"`
-	MemorySwappiness     int         `json:"memory-swappiness" yaml:"memory-swappiness"`
+	MemorySwappiness     OptInt      `json:"memory-swappiness" yaml:"memory-swappiness"`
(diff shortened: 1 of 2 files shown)


crane/container.go
```diff
@@ -103,7 +103,7 @@ type RunParameters struct {
 	RawMemory            string      `json:"memory" yaml:"memory"`
 	RawMemoryReservation string      `json:"memory-reservation" yaml:"memory-reservation"`
 	RawMemorySwap        string      `json:"memory-swap" yaml:"memory-swap"`
-	MemorySwappiness     int         `json:"memory-swappiness" yaml:"memory-swappiness"`
+	MemorySwappiness     OptInt      `json:"memory-swappiness" yaml:"memory-swappiness"`
```
bjaglin: we are loosing type checks with that unfortunately...

michaelsauter:
Thanks a lot for this! I didn't test on earlier Docker versions either. The whole test setup is not the best :(

Anyway. Maybe we should just raise the minimum Docker version more frequently, otherwise this will become unmanageable.

michaelsauter:
I'll release this as 2.3.1 tonight.

bjaglin:
OK, thanks, ping me if I can help for the release.

michaelsauter:
Thanks - released it now: https://github.com/michaelsauter/crane/releases/tag/v2.3.1

PR #247, opened 2015-12-19
configuration: more flexible env/label

bjaglin:
* object syntax was not supported in JSON
* int values were not supported for mapping syntax in YAML
* fail verbosely on unknown types
* added tests


michaelsauter:
Nice!

PR #248, opened 2015-12-21
Top-level networks and volumes

michaelsauter:


michaelsauter:
@bjaglin Here's a first working version. What do you think?

Two notes:
* The implementation is a bit repetitive, possibly due to my limited knowledge of Go. On the other hand, I think it's easy to understand so I'm relatively fine with it
* Currently there are no options for networks and volumes. That makes the configuration a bit awkard:
```yaml
networks:
  bar: {}
```

crane/network.go
```diff
@@ -0,0 +1,33 @@
...
+	Name() string
+	Create()
+	Exists() bool
+}
+
+type network struct {
+	RawName string
+}
+
+func (n *network) Name() string {
```
bjaglin: should the prefix flag/env variable also apply as namespace for networks/volumes? I'd be tempted to say yes

crane/container.go
```diff
@@ -504,15 +519,25 @@ func (r RunParameters) Volume() []string {
...
+		volumes = append(volumes, strings.Join(parts, ":"))
 	}
 	return volumes
 }
 
+func (r RunParameters) VolumeSources() []string {
+	volumes := r.Volume()
+	var volumeSources []string
+	for _, volume := range volumes {
+		parts := strings.Split(volume, ":")
```
bjaglin: maybe name `parts[0]` as `source`?

crane/config.go
```diff
@@ -229,6 +268,25 @@ func (c *config) initialize() {
...
 	}
+
+	c.setNetworkMap()
+	c.setVolumeMap()
+}
+
+func (c *config) setNetworkMap() {
+	c.networkMap = make(map[string]Network)
+	for rawName, network := range c.RawNetworks {
+		network.RawName = rawName
```
bjaglin: I guess better nil handling here would allow
```
networks:
  foo:
```

to be valid without the `{}` to force the allocation of an object when unmarshalling

crane/config.go
```diff
@@ -31,7 +36,11 @@ type config struct {
 	RawContainerMap map[string]*container `json:"containers" yaml:"containers"`
 	RawGroups       map[string][]string   `json:"groups" yaml:"groups"`
 	RawHooksMap     map[string]hooks      `json:"hooks" yaml:"hooks"`
+	RawNetworks     map[string]*network   `json:"networks" yaml:"networks"`
```
bjaglin: `RawNetworkMap` ?

crane/container.go
```diff
@@ -51,16 +51,17 @@ type ContainerInfo interface {
...
-	RawRun    RunParameters   `json:"run" yaml:"run"`
-	RawRm     RmParameters    `json:"rm" yaml:"rm"`
-	RawStart  StartParameters `json:"start" yaml:"start"`
-	RawExec   ExecParameters  `json:"exec" yaml:"exec"`
-	hooks     hooks
+	id          string
+	RawName     string
+	RawUnique   bool            `json:"unique" yaml:"unique"`
+	RawImage    string          `json:"image" yaml:"image"`
+	RawRequires []string        `json:"requires" yaml:"requires"`
```
bjaglin: did you consider having this somewhere else? I can't find of a better place, but it does feel strange there as most of the other values are first-class Docker citizens

crane/unit_of_work.go
```diff
@@ -289,3 +293,55 @@ func (uow *UnitOfWork) ensureInRequireStarted(name string) {
...
+			if includes(volumes, volumeSource) {
+				required = append(required, cfg.Volume(volumeSource))
+			}
+		}
+	}
+	return required
+}
+
+func (uow *UnitOfWork) prepareNetworks() {
+	for _, n := range uow.requiredNetworks() {
```
bjaglin: Maybe these "required" slices should be exposed so that the debug "Command will be applied to" can reference them as well?

bjaglin:
Looks like a solid base! About your 2 notes:

* I think it's just fine the way it is, it's verbose but readable
* If the empty `{}` can be removed as discussed in https://github.com/michaelsauter/crane/pull/248#discussion_r48208082, I think it's fine... I guess the feature can be labeled as experimental so we can always break the format in case it doesn't work out for people in practive.

crane/container.go
michaelsauter: I think it's actually fine here, `unique` is already a non-Docker value. So is `image`, if you think about it - it sort of belongs into the run config, but that has no option `image`. So I would say non-Docker keys at the container level are okay - as there is no equivalent "command" on the Docker side.

crane/config.go
michaelsauter: You're right. I decided to standardize them according to the fields in the config: e66b169cd6e2bcd90c175136f85c3e12876899dc

crane/config.go
michaelsauter: Ah right, implemented in cb8c09b38b8981609a8b99c0df0d006d5e3438db.

crane/network.go
michaelsauter: Sounds like it should be that way, yes. Unfortunately, implementation is not straight-forward. So maybe we can merge this without it and make that change later?

crane/unit_of_work.go
michaelsauter: Good point, added in f27b9ae.

michaelsauter:
Thanks for the great feedback! I addressed most of your comments.

The only major remaining thing is the prefix ...

And yes, integration tests would be great and are desperately needed ... but this PR won't introduce them ;)

bjaglin:
:+1:

bjaglin:
Looks like Compose has the same model in https://github.com/docker/compose/pull/2564. Maybe some documentation and this could be shipped as experimental?

michaelsauter:
@bjaglin Haha, just in time. I added prefix support, an entry to the changelog and some information in the readme. What do you think?

README.md
```diff
@@ -273,6 +276,49 @@ feature is experimental, which means it can be changed or even removed in every
 minor version update.
 
 
+### Networking
+Docker networks are support via the top-level config `networks`, which does not
```
bjaglin: `supported`

README.md
```diff
@@ -273,6 +276,49 @@ feature is experimental, which means it can be changed or even removed in every
...
+    run:
+      net: qux
+  bar:
+    run:
+      net: qux
+networks:
+  qux:
+```
+
+This feature is experimental, which means it can be changed or even removed in
```
bjaglin: line-wrapped on purpose?

README.md
```diff
@@ -273,6 +276,49 @@ feature is experimental, which means it can be changed or even removed in every
...
+every minor version update.
+
+
+### Volumes
+Docker volumes are support via the top-level config `volumes`, which does not
```
bjaglin: `supported`

bjaglin:
LGTM apart from the doc cosmetic comments

Issue #51, opened 2014-06-19
Escaping `$` in configuration values


bjaglin:
Since all values in the `run` configuration are expanded via `os.ExpandEnv()`, I haven't found a way to escape `$` so that it's not expanded.

A typical use case for having non-expanded `$` is the usage within the `CMD` passed to a container of the environment variables such as `$API_PORT_8080_TCP_ADDR,` injected by `docker --link`, which *must* be evaluated on the container and not the host.

Looking at http://golang.org/src/pkg/os/env.go?s=379:436#L3, I don't see any proper way to do it. My current workaround is to define an environment variable `D=$`, and use `${D}`... 

Any other idea? Could the YAML parsing step somehow help out, to differentiate between single and double quotes in values, and therefore applying the expansion only to double-quoted values?

PR #249, opened 2015-12-23
support $$ as a way to escape $

bjaglin:
Fixes https://github.com/michaelsauter/crane/issues/51

docker-compose has the [same behavior](https://docs.docker.com/compose/compose-file/#variable-substitution) as of 1.5.0. I am not particularly proud of the implementation, but it does the job and removes this annoying limitation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Add escape sequence for `$` in configuration
+  `$$` now expands to `$` in all string configuration values.
+  _@bjaglin_
+
 * Add support for Docker networks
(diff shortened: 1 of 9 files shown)


michaelsauter:
Looks good.

Could you update this once the networks/volumes branch is merged?

I think then it's time for another release - probably after Christmas but before the new year!

bjaglin:
I'll rebase and add an entry in the changelog once the other PR is merged.

michaelsauter:
@bjaglin OK, if you could rebase that would be great now that the networks and volumes stuff is merged. Would be great to add a comment to the changeling too.

bjaglin:
@michaelsauter done

PR #250, opened 2015-12-24
support multiple exclusion references & improve doc

bjaglin:
Somehwat related no-op refactoring: leverage https://godoc.org/github.com/alecthomas/kingpin#Application.DefaultEnvars to automatically use env vars as defaults, which has the (good?) side effect all flags can now be set via env variable.

### DIFF
--- a/crane/cli.go
+++ b/crane/cli.go
@@ -25,8 +25,8 @@ var (
 	).Short('p').OverrideDefaultFromEnvar("CRANE_PREFIX").String()
 	excludeFlag = app.Flag(
 		"exclude",
-		"Exclude group or container",
-	).Short('e').OverrideDefaultFromEnvar("CRANE_EXCLUDE").String()
+		"Exclude group or container. Can be repeated.",
+	).Short('e').OverrideDefaultFromEnvar("CRANE_EXCLUDE").PlaceHolder("container|group").Strings()
 	tagFlag = app.Flag(
 		"tag",
 		"Override image tags.",
@@ -227,11 +227,11 @@ func commandAction(targetFlag string, wrapped func(unitOfWork *UnitOfWork), migh
 	wrapped(unitOfWork)
 }
 
-func excludedContainers(flag string) []string {
-	if len(flag) > 0 {
-		return cfg.ContainersForReference(flag)
+func excludedContainers(excludedReference []string) (containers []string) {
+	for _, reference := range excludedReference {
+		containers = append(containers, cfg.ContainersForReference(reference)...)
 	}
-	return []string{}
+	return containers
 }
 
 func runCli() {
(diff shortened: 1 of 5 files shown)


crane/cli.go
```diff
@@ -25,8 +25,8 @@ var (
 	).Short('p').OverrideDefaultFromEnvar("CRANE_PREFIX").String()
 	excludeFlag = app.Flag(
 		"exclude",
-		"Exclude group or container",
-	).Short('e').OverrideDefaultFromEnvar("CRANE_EXCLUDE").String()
+		"Exclude group or container. Can be repeated.",
+	).Short('e').OverrideDefaultFromEnvar("CRANE_EXCLUDE").PlaceHolder("container|group").Strings()
```
bjaglin: FYI, kingpin doesn't support multiple values from an env variable, so the repeated feature will only be available via the CLI flag

crane/cli.go
```diff
@@ -216,11 +216,11 @@ func commandAction(targetFlag string, wrapped func(unitOfWork *UnitOfWork), migh
...
 
-func excludedContainers(flag string) []string {
-	if len(flag) > 0 {
-		return cfg.ContainersForReference(flag)
+func excludedContainers(excludedReference []string) (containers []string) {
+	for _, reference := range excludedReference {
+		containers = append(containers, cfg.ContainersForReference(reference)...)
 	}
-	return []string{}
+	return containers
```
bjaglin: we might have duplicates here, but I guess it's fine?

crane/cli.go
```diff
@@ -12,21 +12,20 @@ var cfg Config
...
 		"prefix",
 		"Container prefix.",
-	).Short('p').OverrideDefaultFromEnvar("CRANE_PREFIX").String()
+	).Short('p').String()
 	excludeFlag = app.Flag(
 		"exclude",
-		"Exclude group or container",
-	).Short('e').OverrideDefaultFromEnvar("CRANE_EXCLUDE").String()
+		"Exclude group or container. Can be repeated.",
+	).Short('e').PlaceHolder("container|group").Strings()
```
bjaglin: FYI, kingpin doesn't support multiple values from an env variable, so the repeated feature will only be available via the CLI flag

bjaglin: Correction: https://github.com/alecthomas/kingpin/pull/93

crane/cli.go
michaelsauter: Should be. You can use the `includes` function to have only unique containers in here.

crane/cli.go
michaelsauter: Will do.

bjaglin:
rebased and added changelog entries

michaelsauter:
:+1:

PR #251, opened 2015-12-29
Parallelize pull

bjaglin:
When lifting a large stack (>20 components), the sequential pulls take a long time, especially with a (private) registry v1. Running them in parallel can cut down drastically the runtime.

This is a WIP to get your opinion on this, the overall concurrency is in place, but what's remaining is proper output and error handling. It shouldn't take long but I want to get your feedback before going any further.

### DIFF
--- a/deps
+++ b/deps
@@ -1,5 +1,5 @@
 github.com/bjaglin/multiplexio 7477705f395a7cc682ae2453b04377e9f4984779
 github.com/fatih/color f19a133fbf02ea4974db468b07328745a8840da7
 github.com/flynn/go-shlex 70644ac2a65dbf1691ce00c209d185163a14edc6
-github.com/alecthomas/kingpin 153fd4a65a216bd58f5ab9b044c6a8d67c401d78
+github.com/alecthomas/kingpin 501bcd10bbdf0c9f970e6392a54c1f2add2721d7
 gopkg.in/v2/yaml 5d6f7e02b7cdad63b06ab3877915532cd30073b4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+* Expose level of concurrency when provisioning
+
+  `lift` and `provision` commands can now be sped-up by passing
+  a custom parallelism level via `--parallel`/`-l`.
+  _@bjaglin_
+
 ## 2.4.0 (2015-12-30)
 
 * Fix a few minor configuration parsing problems when using maps.
--- a/README.md
+++ b/README.md
@@ -381,6 +381,15 @@ The following hooks are currently available:
 Every hook will have the name of the container for which this hook runs available as the environment variable `CRANE_HOOKED_CONTAINER`.
 
 
+### Parallism
+By default, Crane executes all commands sequentially. However, you might want
+to increase the level of parallelism for network-heavy operations, in order to
+cut down the overall run time. The `--parallel`/`-l` flag allows you to
+specify the level of parallelism for commands where network can be a
+bottleneck (namely `provision` and `lift`). Passing a value of 0 effectively
+disable throttling, which means that all provisioning will be done in parallel.
+
+
 ### Container Prefixes
 It is possible to prefix containers with a global `--prefix` flag, which is just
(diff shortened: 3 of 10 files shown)


michaelsauter:
Ah nice! Finally some stuff where Go shines ;)
Generally I think I'm fine with this. However, it seems to me "lift" is the wrong place for this - shouldn't we have this in the "provision" part? Or am I missing something?

Apart from that, one minor change would be to rename from e.g. `GetCommandsOut` to just `CommandsOut` since that seems to be a Go convention. `SetCommandsOut` can stay as is.

bjaglin:
Adapting `provision` to pull in parallel will be easy, the tricky part was the concurrency - basically when you lift, you want to start containers as soon as the corresponding pull is ready (following the order of course). That's a nice-to-have (we could just wait for the slowest pull) that doesn't really bring anything with a simple `provision` so I wanted to start with the hard part first. So I guess I will end up adding the same flag in the `provision` command. By the way `-p` is already taken, so I was thinking about `--parallel/-l`, any opinion? 

About the output redirection: when using `--verbose`, the color output is racy because of a [singleton](https://github.com/michaelsauter/crane/blob/997332f/crane/crane.go#L22) in our color library, but I guess we can live with that.

Duly noted for `CommandsOut`.

michaelsauter:
Okay. Now I get it. Is there a big difference for the user between the two? Starting containers is really quick usually, so I wonder whether there's a big benefit by starting containers the moment they are pulled. Sure you start some containers earlier, but in the end the whole stack shouldn't be up a lot sooner than it would be if you pulled them in parallel, then started them afterwards.

bjaglin:
Fair point, it would be much simpler. Then I'll focus parallelizing `container.Provision()`, dumping respective stdout/stderr as each pull end (and fail hard if any problem), and use that from `unit_of_work.Lift()`.

bjaglin:
I think this is now in reviewable state.

Example with a aarge configuration pulling from an internal registry v1, where all images are already up to date:
```
bjaglin@t440s:~$ time ~/git/golang/bin/crane provision
real    1m44.851s
user    0m0.800s
sys    0m0.180s

bjaglin@t440s:~$ time ~/git/golang/bin/crane provision -l4
real    0m27.702s
user    0m0.872s
sys    0m0.192s

bjaglin@t440s:~$ time ~/git/golang/bin/crane provision -l8
real    0m19.833s
user    0m0.944s
sys    0m0.188s

bjaglin@t440s:~$ time ~/git/golang/bin/crane provision -l0
real    0m12.952s
user    0m1.192s
sys    0m0.292s
```

deps
```diff
@@ -1,5 +1,5 @@
 github.com/bjaglin/multiplexio 7477705f395a7cc682ae2453b04377e9f4984779
 github.com/fatih/color f19a133fbf02ea4974db468b07328745a8840da7
 github.com/flynn/go-shlex 70644ac2a65dbf1691ce00c209d185163a14edc6
-github.com/alecthomas/kingpin 153fd4a65a216bd58f5ab9b044c6a8d67c401d78
+github.com/alecthomas/kingpin 501bcd10bbdf0c9f970e6392a54c1f2add2721d7
```
bjaglin: for https://github.com/alecthomas/kingpin/pull/96

michaelsauter:
Looks good overall! Really like it.

I'm not quite sold on the `0` for doing everything in parallel, but I can't come up with an alternative right now. Let's leave this for a few days and if we can't find anything better, merge it then.

bjaglin:
I started with a boolean flag, but realized there was a need for more granularity to avoid hammering a private registry. Therefore the integer, with semantics similar to `xargs --max-procs` for example. We could simply forbid 0 as a value, but I think there is a use case for unthrottled pulls....

bjaglin:
Could you come up with some alternative? Given the figures I posted earlier and the fact that we lift our stack about 200 times a day on our CI server, this is obviously a very anticipated feature :)

michaelsauter:
No, nothing else I can think of. So, merging :)

michaelsauter:
Released in https://github.com/michaelsauter/crane/releases/tag/v2.5.0.

bjaglin:
:christmas_tree: thanks!

PR #258, opened 2016-01-29
Wait for post-start hook to complete

jesper:
Otherwise, the hook runs in the background, rather than blocking (as was the behavior in Crane v1.x).

### DIFF
--- a/crane/container.go
+++ b/crane/container.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 )
 
 type Container interface {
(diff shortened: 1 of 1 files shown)


bjaglin:
Nice catch. Since you are at it, what about that contract from the README?
> They may interrupt the flow by returning a non-zero status.

michaelsauter:
Ouch. Guess I changed the behaviour when I introduced that without even thinking about it :blush:

jesper:
@bjaglin That seems to remain in-tact after this change is applied. In my local testing, the flow is interrupted and crane halts on a non-zero post-start hook return code.

PR #260, opened 2016-02-17
add push and pull parameters to yml/json

bivas:
ability to specify registry information to pull/push from/to.
Also provide a flag to skip pushing of images (e.g. library ones such as `java:7`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,10 +28,10 @@ continuous integration.
 
 
 ## Installation
-The latest release (2.5.1) can be installed via:
+The latest release (2.5.2) can be installed via:
 
 ```
-bash -c "`curl -sL https://raw.githubusercontent.com/michaelsauter/crane/v2.5.1/download.sh`" && sudo mv crane /usr/local/bin/crane
+bash -c "`curl -sL https://raw.githubusercontent.com/michaelsauter/crane/v2.5.2/download.sh`" && sudo mv crane /usr/local/bin/crane
 ```
 
 Older releases can be found on the
@@ -39,7 +39,7 @@ Older releases can be found on the
 build Crane yourself by using the standard Go toolchain.
 
 Please have a look at the
-[changelog](https://github.com/michaelsauter/crane/blob/v2.5.1/CHANGELOG.md)
+[changelog](https://github.com/michaelsauter/crane/blob/v2.5.2/CHANGELOG.md)
 when upgrading.
 
 Of course, you will need to have Docker (>= 1.6) installed.
--- a/README.md
+++ b/README.md
@@ -28,10 +28,10 @@ continuous integration.
 
 
 ## Installation
-The latest release (2.5.2) can be installed via:
+The latest release (2.6.0) can be installed via:
 
 ```
-bash -c "`curl -sL https://raw.githubusercontent.com/michaelsauter/crane/v2.5.2/download.sh`" && sudo mv crane /usr/local/bin/crane
+bash -c "`curl -sL https://raw.githubusercontent.com/michaelsauter/crane/v2.6.0/download.sh`" && sudo mv crane /usr/local/bin/crane
 ```
 
 Older releases can be found on the
@@ -39,7 +39,7 @@ Older releases can be found on the
 build Crane yourself by using the standard Go toolchain.
 
 Please have a look at the
-[changelog](https://github.com/michaelsauter/crane/blob/v2.5.2/CHANGELOG.md)
+[changelog](https://github.com/michaelsauter/crane/blob/v2.6.0/CHANGELOG.md)
 when upgrading.
 
 Of course, you will need to have Docker (>= 1.6) installed.
--- a/README.md
+++ b/README.md
@@ -285,11 +285,7 @@ in every minor version update.
 
 
 ### Networking
-Docker networks are supported via the top-level config `networks`, which does
-not take additional parameters at this stage. As links cannot be used at the
-same time, container dependencies can be expressed via `requires`. Networks are
-automatically created by Crane when necessary, and never cleaned up. When a
-[prefix](#container-prefixes) is used, it is also applied to the network.
+Docker networks are supported via the top-level config `networks`, which does not take additional parameters at this stage. As links are not strict dependencies for containers attached to a user-defined network (but simply aliases), `requires` can be used instead to indicate that a container must be started for another one to be functional. Networks are automatically created by Crane when necessary, and never cleaned up. When a [prefix](#container-prefixes) is used, it is also applied to the network.
 
 ```
 containers:
--- a/README.md
+++ b/README.md
@@ -27,10 +27,10 @@ continuous integration.
 
 
 ## Installation
-The latest release (2.3.1) can be installed via:
+The latest release (2.4.0) can be installed via:
 
 ```
-bash -c "`curl -sL https://raw.githubusercontent.com/michaelsauter/crane/v2.3.1/download.sh`" && sudo mv crane /usr/local/bin/crane
+bash -c "`curl -sL https://raw.githubusercontent.com/michaelsauter/crane/v2.4.0/download.sh`" && sudo mv crane /usr/local/bin/crane
 ```
 
 Older releases can be found on the
@@ -38,7 +38,7 @@ Older releases can be found on the
 build Crane yourself by using the standard Go toolchain.
 
 Please have a look at the
-[changelog](https://github.com/michaelsauter/crane/blob/v2.3.1/CHANGELOG.md)
+[changelog](https://github.com/michaelsauter/crane/blob/v2.4.0/CHANGELOG.md)
 when upgrading.
 
 Of course, you will need to have Docker (>= 1.6) installed.
--- a/README.md
+++ b/README.md
@@ -381,6 +381,15 @@ The following hooks are currently available:
 Every hook will have the name of the container for which this hook runs available as the environment variable `CRANE_HOOKED_CONTAINER`.
 
 
+### Parallism
+By default, Crane executes all commands sequentially. However, you might want
+to increase the level of parallelism for network-heavy operations, in order to
+cut down the overall run time. The `--parallel`/`-l` flag allows you to
+specify the level of parallelism for commands where network can be a
+bottleneck (namely `provision` and `lift`). Passing a value of 0 effectively
+disable throttling, which means that all provisioning will be done in parallel.
+
+
 ### Container Prefixes
 It is possible to prefix containers with a global `--prefix` flag, which is just
 prepended to the container name. Remember that you will have to provide the same
--- a/README.md
+++ b/README.md
@@ -161,6 +161,13 @@ The map of containers consists of the name of the container mapped to the contai
 	* `context` (string)
 	* `file` (string)
   * `build-arg` (array/mapping) Provide build arguments. Need Docker >= 1.9
+* `push` (object, optional): Parameters related to Docker's `push`.
+  * `skip` (boolean): prevents pushing image when `push` is called
+  * `registry` (string): Docker registry to connect to (will be prefixed to the image name. e.g. `registry.my/user/image`)
+  * `override_user` (string): Override the current image user (e.g. `registry.my/user/image` -> `registry.my/override/image`)
+* `pull` (object, optional): Parameters related to Docker's `pull`.
+  * `registry` (string): Docker registry to connect to (will be prefixed to the image name. e.g. `registry.my/user/image`)
+  * `override_user` (string): Override the current image user (e.g. `registry.my/user/image` -> `registry.my/override/image`)
 * `exec` (object, optional): Parameters mapped to Docker's `exec`.
   * `interactive` (boolean)
   * `tty` (boolean)
--- a/README.md
+++ b/README.md
@@ -28,10 +28,10 @@ continuous integration.
 
 
 ## Installation
-The latest release (2.5.0) can be installed via:
+The latest release (2.5.1) can be installed via:
 
 ```
-bash -c "`curl -sL https://raw.githubusercontent.com/michaelsauter/crane/v2.5.0/download.sh`" && sudo mv crane /usr/local/bin/crane
+bash -c "`curl -sL https://raw.githubusercontent.com/michaelsauter/crane/v2.5.1/download.sh`" && sudo mv crane /usr/local/bin/crane
 ```
 
 Older releases can be found on the
@@ -39,7 +39,7 @@ Older releases can be found on the
 build Crane yourself by using the standard Go toolchain.
 
 Please have a look at the
-[changelog](https://github.com/michaelsauter/crane/blob/v2.5.0/CHANGELOG.md)
+[changelog](https://github.com/michaelsauter/crane/blob/v2.5.1/CHANGELOG.md)
 when upgrading.
 
 Of course, you will need to have Docker (>= 1.6) installed.
--- a/README.md
+++ b/README.md
@@ -455,7 +455,7 @@ As a summary, `&anchor` declares the anchor property, `*alias` is the alias indi
 
 
 ## Copyright & Licensing
-Copyright © 2013-2015 Michael Sauter. See the LICENSE file for details.
+Copyright © 2013-2016 Michael Sauter. See the LICENSE file for details.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -16,22 +16,22 @@ continuous integration.
   * [Groups and Targeting](#groups-and-targeting)
   * [Extending the target](#extending-the-target)
   * [Excluding containers](#excluding-containers)
+  * [Ad hoc commands](#ad-hoc-commands)
   * [Networking](#networking)
   * [Volumes](#volumes)
   * [Hooks](#hooks)
   * [Parallism](#parallism)
   * [Container Prefixes](#container-prefixes)
   * [Override image tag](#override-image-tag)
-  * [Unique Names](#unique-names)
   * [Generate command](#generate-command)
   * [YAML advanced usage](#yaml-advanced-usage)
 
 
 ## Installation
-The latest release (2.6.0) can be installed via:
+The latest release (2.8.0) can be installed via:
 
 ```
-bash -c "`curl -sL https://raw.githubusercontent.com/michaelsauter/crane/v2.6.0/download.sh`" && sudo mv crane /usr/local/bin/crane
+bash -c "`curl -sL https://raw.githubusercontent.com/michaelsauter/crane/v2.8.0/download.sh`" && sudo mv crane /usr/local/bin/crane
 ```
 
 Older releases can be found on the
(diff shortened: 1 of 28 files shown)


michaelsauter:
Thanks for this!

Is there any specific reason why you want to separate the registry from the image field? You can have this as part of the image, e.g. `private-registry.com/user/image:tag`.

Also, is it enough to just always skip official images (by checking if no user is specified)?

bivas:
Skipping official images is kind of trivial - what about `foo/bar`?
The user isn't a library based but I am not the owner of that image. Furthermore, some images, even those owned by me - I wouldn't like to `push` (e.g. temp containers I'm using during development or just locally)

As for the separation of registry from image name: The option here is to `pull` from one registry and `push` to another. The registry separation also gives me the option to keep my configuration line at a sane length. The possibilities are endless (configuration by environment variables even brings additional scenarios we utilize).

I wish to extend this even further and give the ability to change the user when `push`ing and `pull`ing:
1. Use released images when `pull`
2. Modify some content and `push` them to registry as different name (without changing the image name in configuration)
3. Pull my own version of images from registry during development
4. Commit the code changes and CI builds a new version of images.
5. Pull official images again

What do you think?

michaelsauter:
Good point about skipping. 

As for the other points, if you use environment variables already, isn't it more or less the same then if you have it all in one key or spread across multiple keys?

bivas:
My scenario is focused on CI publishing official images while developers work on local version.
When I use environment variables for image name, I can't use `Dockerfile` directive such as 
`FROM company/foo` as the image name is `registry.company.co/company/foo`. 
With my PR, you can state in a single file both the CI and developer views:

* CI builds the images (`crane provision`)
* CI pushes to official registry (`crane push`)
* Developer pulls (`crane pull`). 
  * The registry image is locally tagged as `company/foo` for ease of development.
* `Dockerfile` is pushed to Git. 
* CI servers publish a new set of images (without affecting how `crane.yml` is composed).

If I were to use environment variables, it would have change the image name for me to develop (I could manually re-tag my local images, but I wish to ease the development pain).

michaelsauter:
Hmm. I'm still uncertain about introducing options for commands that correspond to a Docker command that are Crane-specific. E.g. https://docs.docker.com/engine/reference/commandline/pull/ does not have an option `registry`.

Also, I think I'm still not getting your issue, sorry. Wouldn't `image: ${REGISTRY}company/foo` allow you to set `REGISTRY` on CI and leave it empty locally?

bivas:
The issue is not with the image name (as your proposed solution works fine for us), but the need to re-tag pulled images for building local images.
When we do `docker pull registry.company.co/company/foo` the local image is tagged `registry.company.co/company/foo` while dependent `Dockerfile` are starting with `FROM company/foo`.
Using my proposed solution eliminates the need to tag pulled images after using crane. Now, with a single step I pull and re-tag the image.

I know Docker CLI doesn't have a `registry` option for `pull`, but that's what make `crane` different - It makes enterprise scenarios more easy to manage. Also, as I mentioned earlier in my comments, we wish to utilize `crane` to perform even broader Docker development: tagging and pulling images as different username and push them to different registry then the one we pulled from (e.g. copy from one registry server to another (you can look at my local fork on the following branch - https://github.com/bivas/crane/tree/override-user-parameter).

I believe other enterprises will utilize this new behavior. I also prefer using your official repo and not keep my fork too far away.

bjaglin:
Maybe you could post typical `crane.yml`, `Dockerfile`s & commands you would run with that setup?

README.md
```diff
@@ -160,6 +160,11 @@ The map of containers consists of the name of the container mapped to the contai
 * `build` (object, optional): Parameters mapped to Docker's `build`.
 	* `context` (string)
 	* `file` (string)
+* `push` (object, optional): Parameters related to Docker's `push`.
+  * `skip` (boolean): prevents pushing image when `push` is called
```
bjaglin: instead of having a flag per container, I would list all the "pushable" containers in a group, allowing me to push everything I want within a single `git push` without any modification to crane.

crane/container.go
```diff
@@ -1100,10 +1142,31 @@ func (c *container) Logs(follow bool, since string, tail string) (sources []logS
 	return
 }
 
+func (c *container) imageTag(image string, tag string) {
+	fmt.Fprintf(c.CommandsOut(), "Tagging image %s as %s...\n", image, tag)
+	args := []string{"tag", "--force", image, tag}
```
bjaglin: `docker tag --force` is deprecated as of Docker 1.10, and will raise [an error in Docker 1.12](https://github.com/docker/docker/pull/18350#issuecomment-172179799)... Maybe the flag here is passed only for Docker <1.10 to allow for a smooth transition?

README.md
bivas: You mean CLI indication instead of `crane.yml` ? This sounds like a hassle to build a proper command and will hide the actual behavior when developers are just looking at the `yml` file.

bivas:
A sample `crane.yml` might look like:
```
containers:
  os:
    build:
       context: os/
    image: company/os
    push:
       registry: ${DOCKER_USER_REGISTRY}
    pull:
       registry: ${DOCKER_OFFICIAL_REGISTRY}
  service:
    build:
       context: service/
   image: company/service
    push:
       registry: ${DOCKER_USER_REGISTRY}
    pull:
       registry: ${DOCKER_OFFICIAL_REGISTRY}
   requires: [os]

# Omitting other containers, settings and groups
```

`service/Dockerfile` might look like:
```
FROM company/os
ENV SOMETHING 1.1
COPY service.jar /service.jar
```

Now, if I'll pull the os image from the registry, it's local name will be `registry.company.co/company/os`. Modifying and local re-building of `company/service` will force me to re-tag the `os` image as `company/os` for correct reference (or have my `Dockerfile` include the registry information; and you can't user `VAR` in `FROM` statement).
My PR is to allow a git like approach when dealing the container build cycle. You *"fork"* the official repository, make local changes, build the images and can push them to your "fork" without changing the official containers. Once development is completed - the CI servers create the official images and tag them for others to use.

My next PR will enable the use of user forks in `crane` by allowing to override the user to push/pull, but keep the `company/image` tag as described in `crane.yml`. The scenario I'm trying to solve is enterprise centric when CI and development need to see the same picture.

README.md
bjaglin: no groups are defined in `crane.yml`, it's just that this would centralize which containers should be pushed in one place (rather than having a toggle per container).

bivas: I understand. My comment was the process will look like:
```
$ crane build platform
$ crane push platform-push
```
This flow isn't clear enough and bug prone. We have few cases when we wish to skip a specific container push (but not the main scenario).

bjaglin:
Thanks, I understand your use-case now, it's a very advanced usage of Crane!

I must say that I am not using Crane for building, but from my perspective, if @michaelsauter thinks the use-case is worth it, I think your flow would be covered by a more complex `image` definition (while still keeping the existing syntax for backward compatibility and 99% of the usages):
```
containers:
  os:
    build:
       context: os/
    image:
      name: company/os
      push-as: ${DOCKER_USER_REGISTRY}/company/os
      pull-from: ${DOCKER_OFFICIAL_REGISTRY}/company/os
  service:
    build:
       context: service/
   image:
      name: company/service
      push-as: ${DOCKER_USER_REGISTRY}/company/service
      pull-from: ${DOCKER_OFFICIAL_REGISTRY}/company/service
   requires: [os]
```

`image.name` would be required and used for building/running. An empty value for `push-as` or `pull-from` would be the sign that the container should be skipped when the corresponding command is ran.

bivas:
I wouldn't go and change `image` section as other parameters might make it too hard to manage.
I've prepared a post to explain our complete scenario at https://github.com/bivas/crane/blob/override-user-parameter/blog_post.md

michaelsauter:
I actually was thinking about a similar suggestion to @bjaglin's one earlier today.

I really appreciate your PR and am thinking about it, but something sits not right with me atm. I think one issue is that I'm not sure push/pull is the right level for this. What about implicit pulls (e.g. through `crane run`)? 

I'll read your blog post now.

michaelsauter:
@bivas Thanks for writing up your use case. That helps a lot in understanding what you're trying to do. I don't use Crane for building images either, which explains why I'm hesitant to add build-related things quickly.

Now, what you propose definitely sounds quite interesting, so I'd like to continue this discussion. My main concerns at this stage are:

* The process seems difficult to explain. I wonder if there are some adjustments we can make so the concept is easier to grasp?
* As noted earlier, I'm not sure the adjustments are made at the correct location. Could I ask you to think about alternative ideas such as @bjaglin's suggestion and my comments?

bivas:
@michaelsauter @bjaglin Thank you both for commenting and reviewing.

As replied to @bjaglin, I don't think extending the `image` section is the right direction, but I'm willing to change the `pull`/`push` sections to something not so `docker` CLI oriented. I do understand I'm taking `crane` to a different direction. I think `crane` is different than `docker-compose` or other tools in the sense it can be expended to accommodate an advanced usage such as ours (with ease, I might add).

What other adjustments should I provide to better explain our flow? I'm willing to add as much documentation as needed. I do understand this is a very advanced usage and we're building a completely development flow to overcome a lot of issues we found with `docker` flows.

Would you like me to merge both PRs to a single PR so we can try and merge them both to `crane`?

michaelsauter:
@bivas Yes it might help to have the full concept presented in one PR to better understand it.

My idea (though I haven't thought it through) was to maybe add another top-level key called `registry`, which then can hold separate values for push/pull etc. That way we could leave `image` as-is, and augment it if `registry` is defined.

bivas:
Hmmmm... can't seem to figure out why TravisCI fail :(

crane/crane.go
```diff
@@ -92,6 +94,19 @@ func checkDockerClient() {
 			printErrorf("Unsupported client version! Please upgrade to Docker %v or later.\n", intJoin(requiredDockerVersion, "."))
 		}
 	}
+	actualDockerVersion = versions
+}
+
+func validateDockerClientAbove(expectedVersions []int) bool {
```
bjaglin: I am not sure this is really required, it implies a lot of glue code for little value. Besides, it should be a separate PR I think.

bivas: I added it to change behavior on breaking changes (e.g. `docker tag --force`) which should create alternate commands.

bjaglin: Apologies, I had forgotten [my own comment](https://github.com/michaelsauter/crane/pull/260#discussion_r53839650) :( It does introduce some complexity though... any opinion @michaelsauter ?

michaelsauter: Hmm. I'm actually not against raising the minimum required Docker version quite agressively. If you're stuck on an old Docker version, you can still use an old Crane version.

What's the minimum version for `docker tag` without `force` to work as expected?

bivas: IMHO, Docker 1.6 is a good base version. I think enabling features based on their API availability is better for users with multiple docker engine version (which, in large deployments, is a reasonable scenario)

michaelsauter:
@bivas Have you thought about introducing a new top-level key? What do you think?

bivas:
I was thinking that renaming the current top-level keys I've introduced to `pull-config` and `push-config` might be sufficient. I guess you're still not completely comfortable with our use case.
I think @bjaglin proposal of having a `registry` top-level key is also a good direction (`skip` must be available to handle community images). e.g.
```
containers:
  foo:
    image: company/foo
    registry:
       pull-from: ${DOCKER_OFFICIAL_REGISTRY}/company/foo
       push-as: ${DOCKER_USER_REGISTRY}/${USER}/foo
       skip: false
```

What do you think?

michaelsauter:
@bivas Thanks for your suggestion.

I think I have a pretty good understanding now what I think would fit well with Crane. Here's my proposal, based on the suggestions so far:

```
containers:
  foo:
    image: company/foo
    registry:
       pull: ${DOCKER_OFFICIAL_REGISTRY}/company/foo
       push: ${DOCKER_USER_REGISTRY}/${USER}/foo
```

* New, optional `registry` key. If present, sets the registry part of the image. Otherwise, `image` contains the registry
* 2 subkeys `pull` and `push`. They are used in `crane pull` and `crane push`, respectively. The registry in `pull` is also used when lifting/running a container with an image that is not present yet (does that make sense? Not so sure about this part)
* Official images are skipped automatically when pushing
* Custom image cannot be skipped via the config. However it is trivial to add a `pushable` group and do `crane push pushable` if this is really needed.

This is what I think would work well. Thoughts @bjaglin @bivas?

bjaglin:
I think it's a bit strange to have image references under `registry.push/pull`. I do understand there is a need for customizing the namespace/library, but this is strictly speaking not part of the registry endpoint. That's an issue that my last proposal didn't have since it was changing the image configuration (no concept of registry, since it doesn't exist as a standalone thing in Docker CLI either). I don't see anything wrong with extending `image` semantics, as this key/field does not exist as such in the Docker CLI.

That said, since I don't use Crane for building, my opinion is not worth a lot.

bivas:
@michaelsauter one important note - you simply can't figure out if an image is library or community based on its name. A user can tag her image in any name she wish (even overriding an official name). This is why `skip` flag is a must for dealing with image pushing.
I'm not in favor of pushing/pulling images based on groups - as this is easy to break (missing important image in one of the groups; running the wrong CLI command, etc).

I'm well aware that registry doesn't correlates with docker CLI reference - but we're giving an advance scenario that a lot of enterprises are dealing with (usually with `sed` and scripts). I believe `crane` has the best approach of supporting these features (plus - `crane` is really easy to extend - kudos!).

We utilize crane for building, publishing and providing a complete development and test flows for our teams. We also wish to eliminate all kind of script and environment magic and have everything configured in our `crane.yml` files.

@bjaglin extending `image` section to an object will create a breaking change. Or it will introduce a very complex code to maintain and support this change. I do agree it's a more suitable place to configure registry override values.

michaelsauter:
Thanks for the feedback. Guess my suggestion wasn't right just yet. Maybe it's better to not configure a `registry` key, but a `repository`?

Example:
```
containers:
  foo:
    image: foo
    repository:
       pull: ${DOCKER_OFFICIAL_REGISTRY}/company
       push: ${DOCKER_USER_REGISTRY}/${USER}
```

Now we only have the image name in the `image` key, and only repository (in this case with registry prefix) in the `repository` key.

About skipping: While it is true that you can name your image anything you want, I always thought you can't push any image name. Assume you name your image `ruby`, which is also an official image. You can't do `docker push ruby`. So - wouldn't it be safe to skip images named like that? If you specify a repository however, you can push the image of course, since the full image name is something like `private-registry.com/user/ruby`.
Even if that solution isn't quite what you want at this stage, I would propose we try and move forward with this PR without explicit skipping in the configuration. That is an add-on that can always be added later if experience shows it's really needed.

bivas:
regarding `docker push ruby` - you're correct. But nothing says I have to have a user part in my image name: `registry.company.co/ruby` is perfectly fine. Plus, how can you tell if the image is community or from my private registry? e.g. `bivas/gradle` exists in docker HUB but `bivas/service` doesn't.
That's why `skip` is such an important feature.

As for the `registry` vs. `repository` - I don't see the difference. Why not clearly show the intent (which is registry in the case of docker)?

michaelsauter:
My new suggestion fixes 2 things:

* The new key does not contain the image name anymore. This makes more sense, as `image` should still be respected. The image name itself does not need to be repeated as it is always going to be the same (as is the case with GitHub forks)
* The new key also sets the repository part of the image, so `registry` is misleading. Also, it would allow users to just modify the repository, while pushing to the same registry (as is the case with GitHub forks).

So it's not a major difference, but I think it makes slightly more sense and is more consistent.

bivas:
How do you handle the following scenario:
```
containers:
  service:
    image: company/foo
```
where I wish to override `company` with `${USER}`? @bjaglin suggestion of `pull-from`/`push-as` is better fit for this scenario (or my original suggestion of `override_user`).
As for the semantics - in the case of `registry.company.co/company/foo`
- `registry.company.co` is the registry part
- `company/foo` is the repository part
- `company` is the user part
- `foo` is the image name in the `company` repository

So the correct key name should be `registry` as we only change that part. The `override_user` option I added change the `repository` part - but still within the same registry.

michaelsauter:
Oh yes, sorry. I always confuse repository and user, not really sure why. So think of my suggestion as key name `user` not `repository`. However, the term `user` is very broad and might not be a good candidate for a key.

michaelsauter:
Let's get this sorted eventually. One issue I noticed with my suggestion is that it'll be difficult to determine the image name, if you don't include the user part of it. Therefore, the best approach right now would be @bjaglin suggestion I think (keeping BC with current image definition).

```
image:
      name: company/os
      push-as: ${DOCKER_USER_REGISTRY}/company/os
      pull-from: ${DOCKER_OFFICIAL_REGISTRY}/${USER}/os
```
I'm not to fond of the key names, because they are not the same. However I don't have a better alternative, I was considering `push-name`/`pull-name` or `push-as/pull-as` or simply `push/pull`. But unsure if any of those are better.

As I said earlier, I want to keep skipping out for now and consider it afterwards.

bjaglin:
I am not very fan of the suffixes I had suggested either, but I wanted to emphasize the source/target nature of the flags... But anyway, looking back at this now, I think `pull`/`push` would probably be enough. However, if we go that way, maybe `name` is unclear and feels "different".

What about calling it `build` and make it optional? So all flags in the "complex" syntax of `image` would be optional, but at least two would be required (with one you could achieve the same with the regular string syntax). Unset ones will default to: the value `build` if it's set, or `pull` if it's set or `push` if it's set.

```
image:
      build: company/os
      push: ${DOCKER_USER_REGISTRY}/company/os
      pull: ${DOCKER_OFFICIAL_REGISTRY}/${USER}/os
```

```
image:
      build: company/os
      push: ${DOCKER_USER_REGISTRY}/company/os
```

image:
      build: company/os
      pull: ${DOCKER_OFFICIAL_REGISTRY}/${USER}/os
```

bivas:
@bjaglin Having `image` be both value and object is maintenance hell. It might a good direction but will break crane API and I would like to avoid it (and keep backward compatibility)

@michaelsauter As mention earlier - *skip* is a must for our scenarios. The option to pull/push with the same `crane` command and without groups is our main goal. As for the property to use - I think `pull-from` and `push-as` are preferable to just `pull` and `push`. Note on my comment to @bjaglin regarding breaking `image` structure definition. 

If you both don't wish to introduce something you don't usually use `crane` for (building, pull, push) - why break `image` section? I think going back to add a new configuration section is something we can settle as good for both worlds.  

BTW - my initial proposal, of having a new property set keeps the API with backward compatibility and doesn't introduce any code hacks. I agree that having these properties located under some `registry` or some other section might be better.

michaelsauter:
Phew. I'm closing this now as there hasn't been any activity for a while. It seems like we're too far apart to get this in. I hope you understand.

Issue #138, opened 2015-01-27
Investigate vendoring options


michaelsauter:
I want to introduce some dependency management, and https://github.com/gophergala/nut looks perfect so far.

michaelsauter:
Still an issue. http://getgb.io looks like it's gaining some traction and I like the general idea of it.

missedone:
GB is quite good and really easy to use from my view, but one thing you might concern is GB is not go gettable, for more detail see discussion: https://github.com/constabulary/gb/issues/284

michaelsauter:
Right. In the meantime, Go has added experimental support for vendoring and that's an option now too. Since there are no pressing issues right now, I'd like to leave this question open for a little longer and see what happens :)

bjaglin:
I guess https://github.com/michaelsauter/crane/blob/master/deps is obsolete?

michaelsauter:
That should be still what I'm building from since I didn't update any dependencies. However, I'm not really using it right now either.

michaelsauter:
Given Go 1.6 is out and has vendor support by default, I intend to use that now.

The plan is to:

1. Upgrade to Go 1.6
2. Add all dependencies as git submodules (I'll manage them locally with https://github.com/dpw/vendetta)
3. Checkout the versions as specified in the `deps` file
4. Delete the `deps` file
5. Publish

bjaglin:
:+1:

PR #263, opened 2016-02-24
Vendor dependencies as submodule

michaelsauter:
Addresses #138.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,3 +25,6 @@
 [submodule "vendor/gopkg.in/v2/yaml"]
 	path = vendor/gopkg.in/v2/yaml
 	url = https://gopkg.in/v2/yaml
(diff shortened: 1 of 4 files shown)


bjaglin:
https://github.com/michaelsauter/crane/blob/master/.travis.yml#L3 1.6?

Also, maybe removing https://github.com/michaelsauter/crane/blob/master/deps in the same commit would help making sure the hashes are the same?

michaelsauter:
Good points - fixed now.

This is also updating all dependencies (I just checked out master). I had a brief look through the commit history in the projects and it seems alright to me. Looks like coloring was apparently broken in Windows, and maybe is fixed now :D

michaelsauter:
Phew. Now tests actually pass on 1.6 ;) I'm going to use the binary from this branch today and then merge and release 2.6.0.

Issue #261, opened 2016-02-24
Links should not be resolved in networked mode


michaelsauter:
When using network and links together, Crane should not trigger actions on linked containers, as the link is just an alias. To run/start related containers, they need to be in `requires`.

PR #266, opened 2016-02-28
Links are strict dependencies only on bridge networks

bjaglin:
Fixes https://github.com/michaelsauter/crane/issues/261 with (hopefully) a more appropriate strategy than https://github.com/michaelsauter/crane/pull/262

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,11 +285,7 @@ in every minor version update.
 
 
 ### Networking
-Docker networks are supported via the top-level config `networks`, which does
-not take additional parameters at this stage. As links cannot be used at the
-same time, container dependencies can be expressed via `requires`. Networks are
-automatically created by Crane when necessary, and never cleaned up. When a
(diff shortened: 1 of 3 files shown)


michaelsauter:
I might miss the obvious, but what's wrong with #262 ? That you only use links as dependencies in bridge mode? I thought the least intrusive change is to only disregard links when requires is set.

bjaglin:
Sorry for not clarifying.

I think the approach of #262 is wrong because even though it was introduced specifically for networks, `requires` semantics are suited for other contexts for indicating a dependency (potentially without a direct dependency), so relying on that for stripping `link` dependencies may introduce false positives.

This PR goes back to the actual requirements of the docker CLI: `--links` are unsupported on `none`/`container:` `net`, strict dependencies on `bridge`, and optional aliases on user-defined networks.

michaelsauter:
I see. Makes sense. Thanks for taking the time to solve this properly! My change was just a quick bugfix ;)

PR #269, opened 2016-03-07
Add build build-arg parameters

bivas:
Add support for:
https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables-build-arg
https://docs.docker.com/engine/reference/builder/#arg

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,7 @@ The map of containers consists of the name of the container mapped to the contai
 * `build` (object, optional): Parameters mapped to Docker's `build`.
 	* `context` (string)
 	* `file` (string)
+  * `build-arg` (array/mapping) Provide build arguments. Need Docker >= 1.9
 * `exec` (object, optional): Parameters mapped to Docker's `exec`.
   * `interactive` (boolean)
   * `tty` (boolean)
--- a/crane/container.go
+++ b/crane/container.go
@@ -70,8 +70,9 @@ type container struct {
 }
 
 type BuildParameters struct {
-	RawContext string `json:"context" yaml:"context"`
-	RawFile    string `json:"file" yaml:"file"`
+	RawContext   string      `json:"context" yaml:"context"`
+	RawFile      string      `json:"file" yaml:"file"`
+	RawBuildArgs interface{} `json:"build-arg" yaml:"build-arg"`
 }
 
 type RunParameters struct {
@@ -307,6 +308,10 @@ func (b BuildParameters) Context() string {
 
 func (b BuildParameters) File() string {
 	return expandEnv(b.RawFile)
+}
+
+func (b BuildParameters) BuildArgs() []string {
+	return sliceOrMap2ExpandedSlice(b.RawBuildArgs)
 }
 
 func (r RunParameters) AddHost() []string {
@@ -1203,6 +1208,10 @@ func (c *container) buildImage(nocache bool) {
 	if len(c.BuildParams().File()) > 0 {
 		args = append(args, "--file="+filepath.FromSlash(c.BuildParams().Context()+"/"+c.BuildParams().File()))
 	}
+	for _, arg := range c.BuildParams().BuildArgs() {
+		args = append(args, "--build-arg", arg)
+	}
+
 	args = append(args, c.BuildParams().Context())
 	executeCommand("docker", args, c.CommandsOut(), c.CommandsErr())
 	executeHook(c.Hooks().PostBuild(), c.ActualName())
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -160,6 +160,7 @@ The map of containers consists of the name of the container mapped to the contai
 * `build` (object, optional): Parameters mapped to Docker's `build`.
 	* `context` (string)
 	* `file` (string)
+    * `build-arg` (array) Provide build arguments
```
bjaglin: It's a recent Docker addition AFAIK, can you state the minimal version?

bjaglin: array/mapping?

bivas:
Updated `README.md` to reflect required docker version (1.9) when using `build-arg`

bivas:
BTW - in my previous PR I added a method the check against specific command/options when parsing `crane.yml`. We can add this helper method to enforce proper usage if the user is trying to use it with lower docker version.

README.md
```diff
@@ -160,6 +160,7 @@ The map of containers consists of the name of the container mapped to the contai
 * `build` (object, optional): Parameters mapped to Docker's `build`.
 	* `context` (string)
 	* `file` (string)
+    	* `build-arg` (array/mapping) Provide build arguments (requires `docker` version 1.9)
```
bjaglin: other params above say `Need Docker >= 1.9`, maybe align for consistency?

crane/container.go
```diff
@@ -12,6 +11,8 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+
+	"github.com/flynn/go-shlex"
```
bjaglin: do you have a special version of `go fmt`? I don't have the same import ordering

crane/container.go
```diff
@@ -309,6 +311,10 @@ func (b BuildParameters) File() string {
 	return expandEnv(b.RawFile)
 }
 
+func (b BuildParameters) Args() []string {
```
bjaglin: `BuildArgs` ? it's redundant but so is the docker flag name, so it's better to be consistent IMHO

crane/container_test.go
```diff
@@ -299,3 +300,11 @@ func TestOptBoolYAML(t *testing.T) {
 	err := yaml.Unmarshal([]byte("OptBool: notaboolean"), &wrapper)
 	assert.Error(t, err)
 }
+
+func TestBuildArgs(t *testing.T) {
+	var c *container
+	// Simple case
```
bjaglin: superfluous

crane/container.go
bivas: the default `go fmt` (no changes or edits)

bjaglin: that seems to be the `goimports` behavior actually

bjaglin:
Thanks for the PR! Can you address the last comment and squash?

michaelsauter:
:+1: from my side too!

bivas:
@bjaglin are you referring to `goimports` or the simple test?

bjaglin:
Well, actually both are unnecessary changes/additions adding a bit of noise to the actual functional change in my opinion. I do remember @michaelsauter mentioned using `goimports` at some point, but if we do, we should do it in a separate commit/PR to keep the history clean.

crane/container_test.go
bivas: To be honest... I copied `TestActualVolume`. It's simple solution of checking the simplest scenario.
If it's an issue - I'll remove the test.

bjaglin: Sorry if I was unclear - the problem is not the test itself, it's the comment here that doesn't add much value because there is no other scenario, and "Simple" doesn't say much about the scenario, therefore I think it's not worth keeping

bivas: Comment removed.

bjaglin:
Can you squash before I merge? Thanks!

bivas:
squashed it...

bjaglin:
I still see 11 commits, with a lot of merges, are you sure you pushed?

bivas:
I see commit hash f5ab613 as the squashed one. 
I guess it wasn't successful. Do you have any command line guide I should follow?

bjaglin:
much better, now you can just amend clean up the commit message which contains a lot of unuseful information, that would be great!

michaelsauter:
Could you also add an entry to the changelog? That would be great. Just start a new headline `## Unreleased` and add an entry similar to the ones for previous releases. Doesn't have to be perfect, I can clean it up when I release ...

bivas:
Should be OK now.

bjaglin:
Thanks for your patience!

PR #271, opened 2016-03-08
add --net-alias= to docker run flags

dreamcat4:
This PR is for missing docker run flag `--net-alias=`

It is a part of the new docker networking DNS features. As found in docker-engine v1.10.x. Here is the docker docs where it explains the usage scenario for the `--net-alias` flag in respect to new internal dns server:

https://docs.docker.com/engine/userguide/networking/configure-dns/

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ The map of containers consists of the name of the container mapped to the contai
 	* `memory-swap` (string)
 	* `memory-swappiness` (int) Need Docker >= 1.8
 	* `net` (string) The `container:id` syntax is not supported, use `container:name` if you want to reuse another container network stack.
+	* `net-alias` (string)
 	* `oom-kill-disable` (bool) Need Docker >= 1.7
 	* `pid` (string)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -133,6 +133,7 @@ The map of containers consists of the name of the container mapped to the contai
 	* `memory-swap` (string)
 	* `memory-swappiness` (int) Need Docker >= 1.8
 	* `net` (string) The `container:id` syntax is not supported, use `container:name` if you want to reuse another container network stack.
+	* `net-alias` (string)
```
bjaglin: * should be an array
* `Need Docker >= 1.10`

bjaglin:
I have a branch almost ready with all new `docker run` 1.10 flags, which contains this one

bjaglin:
See https://github.com/bjaglin/crane/commit/fc7651a232022353ec73d587dec71e5c6a93ab4a

dreamcat4:
@bjaglin thanks

PR #270, opened 2016-03-08
add --ip= to docker run flags, for #264

dreamcat4:
This PR is for missing docker run flag `--ip=`

https://github.com/michaelsauter/crane/issues/264

PR #272, opened 2016-03-09
`docker run/start/exec` flags catch up

bjaglin:
* `docker run` 1.10 flags (I left `ip` / `ip6` out since it is addressed by https://github.com/michaelsauter/crane/pull/270)
* `docker start/exec` catchup
* fallback on run params for flags common to start & run and exec & run

### DIFF
--- a/crane/container.go
+++ b/crane/container.go
@@ -139,6 +139,7 @@ type RunParameters struct {
 	RawUser              string      `json:"user" yaml:"user"`
 	RawUts               string      `json:"uts" yaml:"uts"`
 	RawVolume            []string    `json:"volume" yaml:"volume"`
+	RawVolumeDriver      string      `json:"volume-driver" yaml:"volume-driver"`
 	RawVolumesFrom       []string    `json:"volumes-from" yaml:"volumes-from"`
 	RawWorkdir           string      `json:"workdir" yaml:"workdir"`
 	RawCmd               interface{} `json:"cmd" yaml:"cmd"`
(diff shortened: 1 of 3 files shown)


bjaglin:
I left `ip` / `ip6` out since it will be addressed by https://github.com/michaelsauter/crane/pull/270

crane/container.go
```diff
@@ -973,7 +1126,14 @@ func (c *container) Start(excluded []string) {
 			if c.StartParams().Attach {
 				args = append(args, "--attach")
 			}
-			if c.StartParams().Interactive {
+			if len(c.StartParams().DetachKeys()) > 0 {
+				args = append(args, "--detach-keys", c.StartParams().DetachKeys())
+			} else {
```
michaelsauter: I'd find `else if` on one line a bit more readable.

michaelsauter:
Whoa. This is awesome! :+1: from my side

crane/container.go
bjaglin: indeed - fixed, amended & force-pushed

crane/container.go
```diff
@@ -1035,12 +1193,28 @@ func (c *container) Exec(cmds []string) {
...
-		if c.ExecParams().Interactive {
+		if c.ExecParams().Detach || c.RunParams().Detach {
+			args = append(args, "--detach")
+		}
+		if len(c.ExecParams().DetachKeys()) > 0 {
+			args = append(args, "--detach-keys", c.ExecParams().DetachKeys())
+		} else if len(c.RunParams().DetachKeys()) > 0 {
+			args = append(args, "--detach-keys", c.RunParams().DetachKeys())
+		}
+		if c.ExecParams().Privileged || c.RunParams().Privileged {
```
bjaglin: That actually hinders the case where the user would want a `true` value in the run block, but an explicit `false` value in the `exec` one. The fix would be to have a `OptBool` but I guess it's not worth it....

michaelsauter: Yea ... I think this is such an edge case that it's not worth it. I'm fine with the proposed behaviour.

bjaglin:
rebased against master

michaelsauter:
Cool. Feel free to merge!

Issue #265, opened 2016-02-28
Implicit unique / ad-hoc behaviour


michaelsauter:
Instead of a pre-defined `unique` key, we could add this behaviour implicitly. 

If a command is passed on the command line, the container gets a timestamp attached to it's name, and the following changes are made to the configuration:

* `publish`, `publish-all` and `detach` are disabled
* `rm` is enabled

Crane would then not care at all about these ad-hoc containers, for example when calling `crane rm` the container would not be removed.

We could have two options `--publish` and `--detach` to override the default behaviour, but I'm not sure it's worth the effort.

The advantage of the proposed change would be:

* No separate configuration key (which name might be difficult to understand)
* It is possible to quickly launch a container with similar configuration without planning for this in advance in the config
* All extra branches in the code related to unique containers can be removed as Crane will only care about the configured containers
* Projects with 2 containers (e.g. a container for a Rails server, and one for ad-hoc rake commands etc.) can simplify the configuration to just one

Since the feature is marked as experimental, this could be released as 2.7.0.

Any thoughts?

bjaglin:
That makes a lot more sense to me than the current somehow-hacky `unique` behavior, and I like the UX :+1:

michaelsauter:
Good to hear :)

PR #274, opened 2016-03-10
Turn unique feature into implicit ad-hoc containers

michaelsauter:
Addresses #265.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,13 +16,13 @@ continuous integration.
   * [Groups and Targeting](#groups-and-targeting)
   * [Extending the target](#extending-the-target)
   * [Excluding containers](#excluding-containers)
+  * [Ad hoc commands](#ad-hoc-commands)
   * [Networking](#networking)
   * [Volumes](#volumes)
   * [Hooks](#hooks)
   * [Parallism](#parallism)
   * [Container Prefixes](#container-prefixes)
   * [Override image tag](#override-image-tag)
-  * [Unique Names](#unique-names)
   * [Generate command](#generate-command)
   * [YAML advanced usage](#yaml-advanced-usage)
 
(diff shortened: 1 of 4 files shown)


crane/container.go
```diff
@@ -1050,12 +1048,16 @@ func (c *container) createArgs(cmds []string, excluded []string) []string {
...
+	if !adHoc {
+		for _, port := range c.RunParams().Publish() {
+			args = append(args, "--publish", port)
+		}
 	}
 	// PublishAll
-	if c.RunParams().PublishAll {
-		args = append(args, "--publish-all")
+	if !adHoc {
+		if c.RunParams().PublishAll {
```
bjaglin: combine in one if ?

crane/container.go
```diff
@@ -933,8 +929,10 @@ func (c *container) createArgs(cmds []string, excluded []string) []string {
 		args = append(args, "--env-file", envFile)
 	}
 	// Expose
-	for _, expose := range c.RunParams().Expose() {
-		args = append(args, "--expose", expose)
+	if !adHoc {
```
bjaglin: I can't see any collision with `expose`, so why blacklist it for adhoc containers?

README.md
```diff
@@ -304,6 +303,15 @@ This feature is experimental, which means it can be changed or even removed
 in every minor version update.
 
 
+### Ad hoc commands
+If you pass a command on the CLI to `lift` or `run`, Crane will add a timestamp
+to the container name (e.g. `foo` will become `foo-1447155694523`), making it
+possible to have multiple containers based on the same Crane config. Those ad
+hoc containers will have `expose`, `publish`, `publish-all` and `detach`
```
bjaglin: `--ip`, `--ip6` should also be ignored since they can collide

bjaglin:
:+1:

crane/container.go
michaelsauter: Oh right. I just saw them mentioned in the CLI docs and thought they worked differently, haven't used that option so far. Now after properly reading the doc I see it's fine. Will undo.

crane/container.go
michaelsauter: Yea, thought it's nicer to read separately.

README.md
michaelsauter: True! I'll add.

michaelsauter:
I'll update and then give it a test run next week at work, if all goes well I'll merge then.

PR #278, opened 2016-03-17
Set public LogSource struct

TeaBough:
Setting `logSource` public allows to mock the `Container` interface from outside the `crane` package.


bjaglin:
@michaelsauter FYI, we are starting to use Crane as a lib in deployment tools (so that we centralize our configuration in `crane.yml`), so you might see more PRs around that showing up over the next weeks

michaelsauter:
Oh nice. I'm totally up for these kind of changes. Thanks!

Issue #277, opened 2016-03-15
Revert run params fallback


michaelsauter:
The run params fallback introduced in #272 causes problems, so it should be reverted.

One issue is that you might define `interactive: true` and `detach: true` in the run params. That works fine, but if you later start the container, the executed command is e.g. `docker start --interactive example`. That attaches to the container, but the desired behaviour would be to detach. Setting `interactive: true` in the run params might be required for certain things in `exec` to work, at least that was the case in earlier Docker versions.

Besides, this fallback introduces some magic that I'm not to comfortable about since the rules are a little unclear.

FYI @bjaglin

bjaglin:
Fair enough, sorry for not thinking though more. Do you want me to do it or you have a PR for it?

michaelsauter:
I don't have prepared anything at the moment. Just wanted to open the issue. If you have time and would like to work on it, highly appreciated. But it's also something I can do before the next release. If you decide to work on it, let me know so we don't do it twice :)

PR #279, opened 2016-03-17
don't fallback on `run` values after all since it can feel unexpected

bjaglin:
Fixes https://github.com/michaelsauter/crane/issues/277


michaelsauter:
Thanks a lot for looking into it!

PR #280, opened 2016-03-18
Set public structs and renamed interfaces

TeaBough:
When using crane as a lib, this will allow to not have to mock systematically the `Config` interface and simply use `Config` struct. To do that the struct must be set public and the interface must be renamed. I've chosen to append `Commander` to the interfaces since those interfaces are supposed to "command" the structs.

### DIFF
--- a/crane/cli.go
+++ b/crane/cli.go
@@ -8,7 +8,7 @@ import (
 	"github.com/alecthomas/kingpin"
 )
 
-var cfg Config
+var cfg ConfigCommander
 var excluded []string
 
 var (
--- a/crane/config.go
+++ b/crane/config.go
@@ -16,7 +16,7 @@ import (
 	"time"
 )
 
-type Config interface {
+type ConfigCommander interface {
 	DependencyMap(excluded []string) map[string]*Dependencies
 	ContainersForReference(reference string) (result []string)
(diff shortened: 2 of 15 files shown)


michaelsauter:
Hmm. I don't like the `Commander` suffix. But I do understand the issue.

How about abbreviations for the structs? E.g. `Vol` (struct) and `Volume` (interface)? 

Do you know how other libraries handle this issue? Or should the interfaces be broken down into smaller ones that speak more specific to where they are used?

TeaBough:
@michaelsauter On second thought, is it really necessary to have those interfaces at all ? Since it's basically just a struct with getter setter afaik.

michaelsauter:
Hmm, not sure. I guess those interfaces are too big as they are right now anyway. Thoughts @bjaglin?

bjaglin:
Not a big fan of the `Commander` suffix either. As for removing the interfaces, I think that's fine, but AFAIK they were introduced to facilitate stubbing in the tests. If that can be solved without interfaces, I am fine with them disappearing.

michaelsauter:
I'll close this since it's quite old and I'm not sold on the suffix. Please re-open with another suggestion if you still care about it.

PR #297, opened 2016-08-17
WIP: Experiment with unison sync

michaelsauter:
Docker for Mac is nice, but the file sharing is extremely slow. Unison sync is nice, but difficult to use and setup.

This PR attempts to fix that problem for Crane users.

Given the following config:
```
containers:
  hello:
    image: ubuntu
    run:
      rm: true
      interactive: true
      tty: true
      volume: 
        - foo:/bar
mac:
  unison-syncs:
    "foo:/bar":
```
Usage is:
```
crane run hello ls -la /bar
```

The code is in early stages, testing is mostly missing.

What's nice about this approach:

* It's fully automated. Crane detects if a volume is configured to be synced via Unison. If so, it runs a container, starts syncing and replaces the `volume` flag with `volumes-from`. Crane also stops syncing if no containers need it anymore.
* If no sync is configured, the default Docker for Mac file sharing will be used.
* Unison syncs can be further customised by specifying `flags` (overriding the default `-auto -batch -confirmbigdel=false -prefer=newer -repeat=watch`), `image` (overriding the default `michaelsauter/unison:2.48.4`), and `uid`/`gid` (overriding the default `0`/`0`)
* Linux binary should not be affected much at all (I guess this can be improved further by build flags).

Latest pre-built binary available at: https://github.com/michaelsauter/crane/releases/tag/experiment-unison-5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+* Optional, fully automated Unison sync for Docker for Mac. Bind-mounts can be
+  marked in the configuration to be synced with Unison. Crane will then manage
+  the lifecycle of the volume container and the Unison sync to it.
+
+  For more information, see [Docker for Mac with Unison sync](https://github.com/michaelsauter/crane#docker-for-mac-with-unison-sync).
+
+  _@michaelsauter_
+
 ## 2.9.1 (2016-08-15)
 
 * Fix tag override for images with no tag specified in the config.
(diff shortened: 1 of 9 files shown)


michaelsauter:
Still a bit unsure which option is best:

1. (Current approach) Starting and stopping the Unison sync for every container automatically with no control by the user. Advantage is that the user does not have to think about syncing it all. 

2. Having explicit `unison start` and `unison stop` commands, but allowing Crane to auto-start the sync. That would mean Crane is simpler. Also, it might make sense to not stop syncing in development every time. Disadvantage is that multiple syncs might be running while you don't need them. We'd probably need to give names to the sync, and we need to have some `unison status` command to see what's running at the moment.

michaelsauter:
Replaced by #298.

PR #311, opened 2016-10-20
Expose the container name as an env variable inside the container

jesper:
It would be useful at times to be able to "introspect" the name of a container without communicating directly via Docker (tedious and security implications). Although it's possible out of the box for a container to know its own ID, finding out the container name is seemingly impossible.

This change automatically inserts a environment variable called CONTAINER_NAME when a "docker run" is being executed.

I considered naming it "CRANE_CONTAINER" or similar (to match CRANE_HOOKED_CONTAINER), however, opted for a more generic name so that the containers themselves do not become "tied down" to crane, even if it's just by naming.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ Following are a list of supported commands and possible options:
 | Command | Maps to | Explanation and Options |
 | ------------- | ----------- | ---------|
 | create      | create           | Creates containers. Any existing containers are removed first. |
-| run         | run              | Runs containers. Any existing containers are removed first. |
+| run         | run              | Runs containers. Any existing containers are removed first. The container's name will be exposed inside the container as the environment variable `CONTAINER_NAME` |
 | rm          | rm               | Removes containers if they exist. Use `--force` to kill running containers first. |
 | start       | start            | Starts containers if not they are not running yet. Non-existing containers are created first. |
 | stop        | stop             | Stops containers if they are running. |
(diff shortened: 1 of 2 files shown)


michaelsauter:
Thanks.

I'm a bit unsure about this since it automatically does something to the container outside of what is visible in the configuration. It seems that you have a specific use case for which this PR provides a solution, but it also applies this "solution" to all containers managed by Crane that don't have the problem.

Maybe you could explain why you need this?

An alternative solution would be to add the container name manually via `--env` config. A bit tedious and is even more complicated if you use the prefix feature. Instead of using `--env`, you could also (mis)use the `--hostname` - I usually set that to the un-prefixed container name.

jesper:
The specific use case for us is passing the container name to gunicorn (http://docs.gunicorn.org/en/stable/settings.html#proc-name) for process naming. Since our setup is heavily template based/dynamic/prefixed, this needs to all be as non-hardcoded as possible.

I understand your hesitation, and indeed felt a bit the same - though unfortunately can't think of a better solution, as adding more config options seems a bit ugly.

The `--hostname` trick is a great idea - thanks for the tip ☺

michaelsauter:
I see. 

Another idea might be to use the entry in `/etc/hosts`:

1. Find out IP of `eth0`
2. Find that IP in `/etc/hosts` and get the associated host name

Of course a lot trickier than what you have with this PR, but it's dynamic and doesn't affect other containers.

michaelsauter:
I'll close this, sorry. I don't see enough of a use case to apply this. I hope you don't mind and that you found another way to achieve what you want.

Issue #124, opened 2014-11-20
Import crane.yml


marmotz:
it would be nice to import others crane.yml files to manage a super environment.

I have a big project with a crane.yml and 4 containers.
I have a little project with a crane.yml and 1 container.

I can boot one project and work on it. But I can't boot all projects in once (or I merge two crane.yml files in a new one)

dreamcat4:
I see this could be useful. Since recently my single `crane.yml` file has grown to be a huge and unwieldy.

The other way \*might\* be to have crane search in subdirs for `crane.yml`. Like file globbing, or a `-r --recursive` option. Perhaps the go language already supports file globbing syntax when specifying the file name (like in ruby). However you would still need crane to load up multiple config files in one go. And sort them out into leaf order before doing so.

Hmm. But then you cannot actually control the load order anymore. It is left up to the gods. Wheras you can with a proper #import directive. Which is actually pretty useful for overriding things (variants, setting private API keys, etc).

gregory:
Would be great to get a feature where we could include another yml file: 


```yml
#db.yml
containers:
    mysql:
        image: mysql
        run:
            env: ["MYSQL_ROOT_PASSWORD=mysecretpassword"]
            detach: true
    memcached:
        image: tutum/memcached
        run:
            detach: true
```

```yml
#crane.yml
include: 
  - ./db.yml
containers:
    apache:
        image: some-apache-image:latest
        run:
            volumes-from: ["app"]
            publish: ["80:80"]
            link: ["mysql:db", "memcached:cache"]
            detach: true
    app:
        dockerfile: app
        image: michaelsauter/app
        run:
            volume: ["app/www:/srv/www:rw"]
            detach: true
```

including db would have the keys from crane.yml merged over the ones of db.yml

In our case, we have couple services with a lot of env variables and the crane file starts to be pretty big...

gregory:
@michaelsauter regarding 
> I really should add something to the readme that states that I want to keep Crane as minimal as possible so I can refer to it all the time :)

I totally agree, though this would not increase the scope of what crane is doing but rather provide us with the ability/flexibility of having minimal & modular crane files. 

I checked again docker-compose yesterday and there is couple things that, IMHO they are doing wrong. There is no way we could get that working in our dev env (SOA).
Turns out our crane.yml started to get pretty big and harder to read/maintain. 

This feature would be pretty useful.

saidimu:
[...]

PR #314, opened 2016-11-13
Implement config override

michaelsauter:


michaelsauter:
Previously attempted in #214, and somewhat related to #124.

Issue #335, opened 2018-03-05
Runtime error if a network is specified on a container, but the network isn't in the networks list


inthroxify:
On Crane v3.3.3 I get a runtime error if a network is specified on a container, but the network isn't in the networks list.

Given the following sample `crane.yml`:
```
services:
  crane:
    image: michaelsauter/golang:1.8
    rm: true
    interactive: true
    tty: true
    volume: ["$GOPATH:/go"]
    workdir: /go/src/github.com/michaelsauter/crane
    cmd: ["bash"]
    networks:
        asplode:

accelerated-mounts:
  crane:
    uid: 1000
    gid: 1000
```

Results in the following runtime error (using dry-run option doesn't affect the error outcome either way, but it saves me from cleaning up a container afterwards):
```
$ go run main.go --dry-run start 
Reading configuration /home/user/gocode/src/github.com/michaelsauter/crane/crane.yml
Command will be applied to: crane
If needed, also creates networks: default

Creating network crane_default ...
--> docker network create crane_default
Running container crane_crane ...
--> docker create --interactive --rm --tty --volume /home/user/gocode:/go --workdir /go/src/github.com/michaelsauter/crane --name crane_crane michaelsauter/golang:1.8 bash
ERROR: runtime error: invalid memory address or nil pointer dereference
exit status 1
```

Good-ish news is that things work expected if the network is specified, of course:
```
services:
  crane:
    image: michaelsauter/golang:1.8
    rm: true
    interactive: true
    tty: true
    volume: ["$GOPATH:/go"]
    workdir: /go/src/github.com/michaelsauter/crane
    cmd: ["bash"]
    networks:
        worksfine:

accelerated-mounts:
  crane:
    uid: 1000
    gid: 1000

networks:
    worksfine:
```
Results in:
```
$ go run main.go --dry-run start 
Reading configuration /home/user/gocode/src/github.com/michaelsauter/crane/crane.yml
Command will be applied to: crane
If needed, also creates networks: worksfine, default

Creating network crane_worksfine ...
--> docker network create crane_worksfine
Creating network crane_default ...
--> docker network create crane_default
Running container crane_crane ...
--> docker create --interactive --rm --tty --volume /home/user/gocode:/go --workdir /go/src/github.com/michaelsauter/crane --name crane_crane michaelsauter/golang:1.8 bash
--> docker network connect --alias crane crane_worksfine crane_crane
--> docker network connect --alias crane crane_default crane_crane
[...]

PR #336, opened 2018-03-05
handle missing network

inthroxify:
Potential fix for #335 but it's not terribly elegant.


michaelsauter:
Thanks a lot! This will help to spot the actual issue a lot easier ...

PR #352, opened 2019-09-27
prevent duplicate when containers are both dependencies & affected

bjaglin:
Hi @michaelsauter, long time no see!

This is a bug that has been biting us for a long time, but we always worked around it by adding containers to the groups to force de-duplication (that would be like adding `b` to the target in the added test case).

Note that this is targeting 2.x as we are still actively using crane 2.x as the main engine of our entire CI (hopefully I'll find a time to write a blog post on https://medium.com/teads-engineering to describe our setup and our journey), and we will most likely never move to 3.x as it will incur dozens if not hundreds of pull requests across the organization. So that would be great if you cut a 2.x release with this so I don't need to setup a mirror internally to build it!

### DIFF
--- a/crane/target_test.go
+++ b/crane/target_test.go
@@ -139,10 +139,14 @@ func TestDeduplicationAll(t *testing.T) {
 	)
 	groups := map[string][]string{
 		"ab": []string{"a", "b", "a"},
+		"ac": []string{"a", "c"},
 	}
 	cfg = &config{containerMap: containerMap, groups: groups}
 	dependencyMap := cfg.DependencyMap()
 
-	target, _ := NewTarget(dependencyMap, "ab+dependencies+affected")
-	assert.Equal(t, []string{"a", "b", "c"}, target.all())
+	target1, _ := NewTarget(dependencyMap, "ab+dependencies+affected")
+	assert.Equal(t, []string{"a", "b", "c"}, target1.all())
+
+	target2, _ := NewTarget(dependencyMap, "ac+dependencies+affected")
+	assert.Equal(t, []string{"a", "b", "c"}, target2.all())
 }
--- a/crane/target.go
+++ b/crane/target.go
@@ -146,7 +146,9 @@ func (t Target) all() []string {
 		all = append(all, name)
 	}
 	for _, name := range t.affected {
-		all = append(all, name)
+		if !includes(all, name) {
+			all = append(all, name)
+		}
 	}
 	sort.Strings(all)
 	return all


crane/target_test.go
```diff
@@ -139,10 +139,14 @@ func TestDeduplicationAll(t *testing.T) {
...
 	cfg = &config{containerMap: containerMap, groups: groups}
 	dependencyMap := cfg.DependencyMap()
 
-	target, _ := NewTarget(dependencyMap, "ab+dependencies+affected")
-	assert.Equal(t, []string{"a", "b", "c"}, target.all())
+	target1, _ := NewTarget(dependencyMap, "ab+dependencies+affected")
+	assert.Equal(t, []string{"a", "b", "c"}, target1.all())
+
+	target2, _ := NewTarget(dependencyMap, "ac+dependencies+affected")
+	assert.Equal(t, []string{"a", "b", "c"}, target2.all())
```
bjaglin: `b` was duplicated before the fix

michaelsauter:
@bjaglin Indeed! I thought you probably dropped working with Crane. In fact, it's so long ago that I had to look through the changelog to remind myself what changes happened from v2 to v3. You're right, quite a lot changed... sorry that the upgrade is hard for you.

I will try to release 2.11.1 with your fix.

michaelsauter:
Hmm sorry, I can't build it anymore. That was pre-modules etc. and I don't have that setup anymore.

What do you think about just maintaining 2.x yourself under your username / org? You can upload the binaries there yourself, then you have pretty much the same setup, but you can also do bugfixes by yourself.

Let me know if that works for you.

dreamcat4:
Hey guys....
So why don't you both register under a new github organization called `crane-orchestration/crane` ? As that already seems to be the name of the website.

Then two you can better manage the admin side of things going forwards. Sorry if I missed a part of the conversation about it.

bjaglin:
@dreamcat4 we figured it out in https://github.com/michaelsauter/crane/pull/353. Out of curiosity, are you also using the 2.x branch?

dreamcat4:
Hey there. No I appear to be on `v3.5.0` now. But in order to do that there were some changes I had to make in my `crane.yml` file. In order to get it working.

So... what can I suggest about the process of upgrading from 2.x to 3.x ?

Well perhaps (like me), when using YAML configs rather than JSON. Then the main stumbling block here is the need to adjust the levels of indentation to result in a correctly formatted adjusted yaml file. Which is not simple to automate with the typical sed command. That would be my guess.

Oh I suppose it would be possible to convert the files over to Json first. Where it could be more easy to make such changes in json. Which is not sensitive to the indentation. Then convert it back to yaml again afterwards. IDK.

The other possible way is [learning how to deal with YAML indentation in awk](https://stackoverflow.com/a/56878445/287510), and / or asking for help from the experts on the #awk channel on freenode.

bjaglin:
@dreamcat4  thanks for the feedback. In my case, the problem is not to update `crane.yml` but the dozens of repos that rely for their integration tests on the 2.x CLI syntax (https://github.com/michaelsauter/crane/blob/master/CHANGELOG.md#300-2017-04-08) and legacy Docker networks. So it's deep, structural change at an organization level.

PR #353, opened 2019-09-29
prepare 2.x release

bjaglin:
Follow-up of https://github.com/michaelsauter/crane/pull/352#issuecomment-536203220: I got `release.sh` up and running locally with this (so that unlocks my need which was to get a linux binary to wrap in our internal Docker image).

I don't think I'll go as far as publishing under my own username/org though. However, I think I am (still) a collaborator on this repo, so if you don't mind, I can take care of crafting that last 2.x release here myself?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+* Fix failure when a command is applied to a group with dependencies & affected
+  containers, and dependencies & affected containers intersect.
+
+  _@bjaglin_
+
 ## 2.11.0 (2016-11-14)
 
 * Allow to override configuration by another configuration file. By default,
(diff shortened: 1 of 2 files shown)


michaelsauter:
Fine by me!

bjaglin:
Done in https://github.com/michaelsauter/crane/releases/tag/v2.11.1, hopefully I didn't forget anything!

bjaglin:
Even though I didn't push the [`latest` tag](https://github.com/michaelsauter/crane/releases/tag/latest), https://github.com/michaelsauter/crane/releases/latest does redirect to my release. This could break the workflow of 3.x users :confused: It looks like there is a fishy way to workaround it, in case it's a concern: https://stackoverflow.com/questions/22822586/swap-latest-release-on-github

PR #361, opened 2021-11-22
Follow up from PR #360

michaelsauter:


dreamcat4:
Thanks so much!

Just tried it here as `/v3@latest`, and it worked perfectly, pulling in the as new`v3.6.1` tag. With `crane version` reporting `v3.6.1`. Which is fantastic

